PR #1, opened 2020-06-30
run: Add wrapper for Cloud Run package

gvso:
We need a wrapper for the Cloud Run package to make testing more manageable. 

As of now, I'm leaving the region as a parameter. In following phases, we can make the tool automatically detect the region of the requested service or retrieve it from environment variables.


pkg/run/wrapper.go
```diff
@@ -0,0 +1,46 @@
...
+	Client *run.APIService
+	Region string
+}
+
+// NewAPIClient initializes an instance of APIService.
+func NewAPIClient(ctx context.Context, region string) (*API, error) {
+	regionalEndpoint := fmt.Sprintf("https://%s-run.googleapis.com/", region)
+	client, err := run.NewService(ctx, option.WithEndpoint(regionalEndpoint))
+	if err != nil {
+		return nil, errors.Wrap(err, "could not initialize a APIService instance")
```
ahmetb, line 29: `could not initialize cloud run client`

pkg/run/mock/runclient.go
```diff
@@ -0,0 +1,24 @@
+package mock
```
ahmetb, line 1: use dir: `internal/run`

PR #4, opened 2020-07-01
rollout: Perform rollout process a step at a time

gvso:
This allows roll forward of a revision one step for each execution of the binary. The process is the core of how rollouts would be managed.


gvso:
I'm sorry for the relatively big change. I wanted to provide something that works when running it. The bulk of the logic is in `SplitTraffic` actually.

The configuration is hardcoded in the `main.go` file. You can change it there if you want to try some other service.

pkg/rollout/revision.go
```diff
@@ -0,0 +1,57 @@
+package rollout
+
+import (
+	"google.golang.org/api/run/v1"
+)
+
+// Annotations name for information related to the rollout.
+const (
+	StableRevisionAnnotation    = "rollout.cloud.run/stableRevision"
```
gvso, line 9: Any ideas of another annotation key? Or is this fine?

pkg/rollout/rollout.go
```diff
@@ -0,0 +1,191 @@
...
+type Rollout struct {
+	RunClient        runapi.Client
+	Config           *config.Config
+	Log              Logger
+	promoteCandidate bool
+}
+
+// Automatic tags.
+const (
+	StableTag    = "stable"
```
gvso, line 29: Do you prefer `StableTag`, `CandidateTag` or `TagStable`, `TagCandidate`?

ahmetb: this is good

pkg/rollout/revision.go
ahmetb: this is good for now.

cmd/operator/main.go
```diff
@@ -14,5 +14,45 @@
...
+			Project: "gtvo-test",
+			Service: "hello",
+			Region:  "us-east1",
+		},
+		Rollout: &config.Rollout{
+			Steps: []int64{5, 30, 80},
+		},
+	}
+	log := log.New()
+
```
ahmetb, line 40: init `ctx := context.Background()`

cmd/operator/main.go
```diff
@@ -14,5 +14,45 @@
...
+		Metadata: &config.Metadata{
+			Project: "gtvo-test",
+			Service: "hello",
+			Region:  "us-east1",
+		},
+		Rollout: &config.Rollout{
+			Steps: []int64{5, 30, 80},
+		},
+	}
+	log := log.New()
```
ahmetb, line 39: add explicit config: if output is a TTY use `logfmt` format
if output is not a TTY (headless execution, deployed) use `json` format
(structured logging)

cmd/operator/main.go
```diff
@@ -14,5 +14,45 @@
...
+
+	client, err := run.NewAPIClient(context.TODO(), config.Metadata.Region)
+	if err != nil {
+		log.Fatal("could not initilize Cloud Run client", err)
+	}
+	roll := rollout.New(client, config, log)
+
+	svc, err := roll.Manage()
```
ahmetb, line 47: don't return svc.

pkg/rollout/revision.go
```diff
@@ -0,0 +1,57 @@
...
+	CandidateRevisionAnnotation = "rollout.cloud.run/candidateRevision"
+)
+
+// StableRevision returns the stable revision of the Cloud Run service.
+func StableRevision(svc *run.Service) string {
```
ahmetb, line 14: two options:

`DetectStableRevisionName`

or:

`type Service struct { *run.Service }` then define StableRevisionName() method on this type.

pkg/rollout/revision.go
```diff
@@ -0,0 +1,57 @@
...
+	if stable == latestRevision {
+		return ""
+	}
+
+	return latestRevision
+}
+
+// detectTrafficHandler scans the service and retrieves a revision with 100%
+// traffic.
+func detectTrafficHandler(svc *run.Service) string {
```
ahmetb, line 48: detectTrafficHandler name doesn't make a lot of sense :)

ahmetb: find100PercentServingRevisionName

pkg/rollout/revision.go
```diff
@@ -0,0 +1,57 @@
...
+	"google.golang.org/api/run/v1"
+)
+
+// Annotations name for information related to the rollout.
+const (
+	StableRevisionAnnotation    = "rollout.cloud.run/stableRevision"
+	CandidateRevisionAnnotation = "rollout.cloud.run/candidateRevision"
+)
+
+// StableRevision returns the stable revision of the Cloud Run service.
```
ahmetb, line 13: add godoc on how does it determine

pkg/rollout/revision.go
```diff
@@ -0,0 +1,57 @@
...
+func StableRevision(svc *run.Service) string {
+	stableRevision, ok := svc.Metadata.Annotations[StableRevisionAnnotation]
```
ahmetb, line 15: detection mechanism should look at `traffic.tag` before annotation.

let's see if we can make annotation more of an "info" to the user

pkg/rollout/revision.go
```diff
@@ -0,0 +1,57 @@
...
+
+import (
+	"google.golang.org/api/run/v1"
+)
+
+// Annotations name for information related to the rollout.
+const (
+	StableRevisionAnnotation    = "rollout.cloud.run/stableRevision"
+	CandidateRevisionAnnotation = "rollout.cloud.run/candidateRevision"
+)
```
ahmetb, line 11: suggestion: introduce a `lastUpdate` annotation.

imagine we see a service 50-50 split, no annotation, no tag. we're stuck, so inform user why.

pkg/rollout/revision_test.go
```diff
@@ -0,0 +1,92 @@
...
+		svc := generateService(opts)
+		stable := rollout.StableRevision(svc)
+
+		assert.Equal(t, test.expected, stable)
+	}
+}
+
+func TestGetCandidateRevision(t *testing.T) {
+	var tests = []struct {
+		lastestReady   string
```
ahmetb, line 67: typo

pkg/rollout/revision_test.go
```diff
@@ -0,0 +1,92 @@
...
+		// Latest Revision is not the same as the stable one.
+		{
+			lastestReady:   "test-002",
+			stableRevision: "test-001",
+			expected:       "test-002",
+		},
+	}
+
+	for _, test := range tests {
+		opts := &ServiceOpts{LatestReadyRevision: test.lastestReady}
```
ahmetb, line 86: use `name` and `t.Run`

pkg/rollout/rollout.go
```diff
@@ -0,0 +1,191 @@
...
+	format "github.com/logrusorgru/aurora"
+	"github.com/pkg/errors"
+	"google.golang.org/api/run/v1"
+)
+
+// Logger describes a logger printing information to the user.
+type Logger interface {
+	Println(args ...interface{})
+	Printf(format string, args ...interface{})
+}
```
ahmetb, line 17: with logrus just use logrus's own interface, don't define an interface.

pkg/rollout/rollout.go
```diff
@@ -0,0 +1,191 @@
...
+	LatestTag    = "latest"
+)
+
+// New returns a new rollout manager.
+func New(client runapi.Client, config *config.Config, log Logger) *Rollout {
+	return &Rollout{
+		RunClient:        client,
+		Config:           config,
+		Log:              log,
+		promoteCandidate: false,
```
ahmetb, line 40: omit defaults

pkg/rollout/rollout.go
```diff
@@ -0,0 +1,191 @@
...
+	Println(args ...interface{})
+	Printf(format string, args ...interface{})
+}
+
+// Rollout is the rollout manager.
+type Rollout struct {
+	RunClient        runapi.Client
+	Config           *config.Config
+	Log              Logger
+	promoteCandidate bool
```
ahmetb, line 24: add a line between configs and state

pkg/rollout/rollout.go
```diff
@@ -0,0 +1,191 @@
...
+	CandidateTag = "candidate"
+	LatestTag    = "latest"
+)
+
+// New returns a new rollout manager.
+func New(client runapi.Client, config *config.Config, log Logger) *Rollout {
+	return &Rollout{
+		RunClient:        client,
+		Config:           config,
+		Log:              log,
```
ahmetb, line 39: initialize logger with fields.

pkg/rollout/rollout.go
ahmetb, line 17: remove, just work with logrus

pkg/rollout/rollout.go
```diff
@@ -0,0 +1,191 @@
...
+		Percent:      percent,
+		Tag:          tag,
+	}
+}
+
+// generateServiceName returns the name of the specified service. It returns the
+// form namespaces/{namespace_id}/services/{service_id}.
+//
+// For Cloud Run (fully managed), the namespace is the project ID or number.
+func generateServiceName(namespace, serviceID string) string {
```
ahmetb, line 189: formalServiceName or just serviceName

pkg/rollout/rollout.go
```diff
@@ -0,0 +1,191 @@
...
+		Config:           config,
+		Log:              log,
+		promoteCandidate: false,
+	}
+}
+
+// Manage handles the gradual rollout.
+func (r *Rollout) Manage() (*run.Service, error) {
+	serviceName := generateServiceName(r.Config.Metadata.Project, r.Config.Metadata.Service)
+	svc, err := r.RunClient.Service(serviceName)
```
ahmetb, line 47: get the project and service name here, don't get API-specific naming convention into this part of code

pkg/rollout/rollout.go
```diff
@@ -0,0 +1,191 @@
...
+	if err != nil {
+		return nil, errors.Wrap(err, "could not get service")
```
ahmetb, line 49: add %q

pkg/rollout/rollout.go
```diff
@@ -0,0 +1,191 @@
...
+	}
+	if svc == nil {
+		return nil, errors.Errorf("service %s does not exist", serviceName)
+	}
+
+	stable := StableRevision(svc)
+	if stable == "" {
+		r.Log.Println("Could not determine stable revision")
```
ahmetb, line 57: with logrus you get levels, use debug level for these.

pkg/rollout/rollout.go
```diff
@@ -0,0 +1,191 @@
...
+	candidateTarget, candidatePercent := r.candidateTraffic(svc, candidate)
+	if candidateTarget == nil {
+		candidatePercent = r.Config.Rollout.Steps[0]
+	} else if candidatePercent == candidateTarget.Percent {
+		// If the traffic share did not change, candidate already handled 100%
+		// and is now ready to become stable.
+		r.promoteCandidate = true
+	}
+
+	var traffic []*run.TrafficTarget
```
ahmetb, line 93: - do not try to modify spec.traffic section, just read it and gather info (candidate, stable, user-specified tags)

- create a new traffic section, with the candidate/stable rules (two rules)

- detect user-specified tags (rules without percentage, and tag != stable or candidate) `{revisionName: x, tag: y}`

- append the rules with user-specified tags into the new traffic section to preserve it

pkg/rollout/rollout.go
```diff
@@ -0,0 +1,191 @@
...
+			continue
+		}
+
+		// When the user introduces a tag manually (UI/gcloud), Cloud Run breaks
+		// all targets with (Name, Tag, Percent) into two different targets
+		// (Name, Tag) and (Name, Percent).
+		// This recovers from this by ignoring the (Name, Tag) target.
+		if (target.Tag == StableTag || target.Tag == CandidateTag) && target.Percent == 0 {
+			continue
+		}
```
ahmetb, line 115: encapsulate in a func like isTagOnlyTrafficRule

pkg/rollout/rollout.go
```diff
@@ -0,0 +1,191 @@
...
+
+	return svc
+}
+
+// candidateTraffic returns the traffic configuration for the candidate and the
+// next traffic share for the candidate.
+func (r *Rollout) candidateTraffic(svc *run.Service, candidate string) (*run.TrafficTarget, int64) {
+	for _, target := range svc.Status.Traffic {
+		if target.RevisionName == candidate && target.Percent > 0 {
+			return target, r.nextCandidateTraffic(target.Percent)
```
ahmetb, line 142: this decision is too deep in the chain, hence not very flexible, move to SplitTraffic

pkg/rollout/rollout.go
```diff
@@ -0,0 +1,191 @@
...
+			return step
+		}
+	}
+
+	return 100
+}
+
+// updateAnnotations updates the annotations to keep some state about the rollout.
+func (r *Rollout) updateAnnotations(svc *run.Service, stable, candidate string) *run.Service {
+	// The candidate has become the stable revision.
```
ahmetb, line 162: `svc.Metadata` can be null, make(map[string]string)

pkg/rollout/rollout.go
ahmetb, line 40: add comment

pkg/rollout/rollout_test.go
```diff
@@ -0,0 +1,418 @@
...
+			}
+		}
+
+		assert.True(t, found, "Expected target traffic not found.", t1)
+	}
+
+}
+
+// equalArray checks that the values of two arrays are equivalent.
+func assertArray(t *testing.T, expected, actual []string) {
```
ahmetb, line 399: golang cmp.Equals

cmd/operator/main.go
gvso: What exactly do you mean `json` format? I know by experience that when deployed in Cloud Run, the logs have the format

```
level=info msg=This is a message for the user project=xyz123 region=us-east1 service=aservice123
```

Isn't that sufficient? Do we have to make it look like a `json` output?

cmd/operator/main.go
gvso: I agree that it might be better not to return a full service to the user. I have two concerns about this, though

1) We still need to return something since `nil` svc means that nothing changed but that there was no errors (e.g. no candidate exists)
2) The returned service is useful for testing that the changes happened as expected. We could solve this by writing a a function that calls `Manage`. Something like:
```
func (r *Rollout) Run(...) (someStruct, error) {
   svc, err := r.Manage(...)
}
```
The Rollout instance would then just call `Run` instead of manage and we can still test that the service was changed as expected. However, I don't know if you think that this is the best approach.

Any thoughts?

pkg/rollout/revision.go
gvso: I'll work on this on a following PR. Is that ok?

cmd/operator/main.go
ahmetb: Most idiomatic way to do this:

`changed, err := roll.RolloutService(svc)`

pkg/rollout/rollout_test.go
```diff
@@ -0,0 +1,376 @@
...
+				Traffic:             test.traffic,
+			}
+			return generateService(opts), nil
+		}
+		client.ReplaceServiceFn = func(namespace, serviceID string, svc *run.Service) (*run.Service, error) {
+			return svc, nil
+		}
+
+		lg := logrus.New()
+		lg.Out = ioutil.Discard
```
ahmetb, line 188: SetOutput

PR #5, opened 2020-07-08
config: Allow auto rollouts using flags for configuration

gvso:
This creates methods in the config package to initialize and validate a configuration based on CLI flags. It also automatically rolls forward every <interval> amount of time (in seconds)

The proposed way to run the program is as follows:

`./cloud-run-release-operator -cli -project=myproject -region=us-east1 -service=mysvc -step=5 -step=30 -step=60 -interval=20`

OR

`./cloud-run-release-operator -cli -project=myproject -region=us-east1 -service=mysvc -steps=5,30,60 -interval=20`

### DIFF
--- a/cmd/operator/main.go
+++ b/cmd/operator/main.go
@@ -29,7 +29,6 @@ import (
 	isatty "github.com/mattn/go-isatty"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
-	log "github.com/sirupsen/logrus"
 )
 
 type stepFlags []int64
@@ -73,6 +72,10 @@ func init() {
 	flag.StringVar(&flStepsString, "steps", "", "define steps in one flag separated by commas (e.g. 5,30,60)")
 	flag.Int64Var(&flInterval, "interval", 0, "the time between each rollout step")
 	flag.Parse()
+}
+
+func main() {
+	logger := logrus.New()
 
 	// -steps flag has precedence over the list of -step flags.
 	if flStepsString != "" {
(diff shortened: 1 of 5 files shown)


pkg/config/config_test.go
```diff
@@ -0,0 +1,145 @@
...
+	tests := []struct {
+		name      string
+		data      []byte
+		cliMode   bool
+		expected  *config.Config
+		shouldErr bool
+	}{
+		{
+			name:    "correct config file",
+			data:    []byte(yaml1),
```
gvso, line 70: Any recommendations here? Is it better to put the YAML string in here instead? Would it look good, especially because YAML is very strict about indentation?

ahmetb: `string` is ok. but yeah the indentation will be off a bit.

cmd/operator/main.go
```diff
@@ -16,46 +16,123 @@ package main
...
-	if err != nil {
-		log.Fatalf("could not initilize Cloud Run client: %v", err)
+	if flCLI {
+		runCLI(logger, flConfigFile)
+	}
+}
+
+func runCLI(logger *logrus.Logger, file string) {
+	cfg := config.WithValues(flProject, flRegion, flService, flSteps, flInterval)
+	if !cfg.IsValid(true) {
```
ahmetb, line 116: I'm not sure why validation is called so late? Can't we validate it when we read them (e.g. in `init()` or as the first thing in `main()`?

cmd/operator/main.go
```diff
@@ -16,46 +16,123 @@ package main
...
+	flag.StringVar(&flRegion, "region", "", "the Cloud Run region where the service is deployed")
+	flag.StringVar(&flService, "service", "", "the service to manage")
+	flag.Var(&flSteps, "step", "a percentage in traffic the candidate should go through")
+	flag.StringVar(&flStepsString, "steps", "", "define steps in one flag separated by commas (e.g. 5,30,60)")
+	flag.Int64Var(&flInterval, "interval", 0, "the time between each rollout step")
+	flag.Parse()
+
+	if !flCLI && flHTTPAddr == "" {
+		log.Fatal("one of -cli or -http-addr must be set")
+	}
```
ahmetb, line 80: might be wiser to validate these in `main()` I think, `init()`s should be as small as possible.

also in `main()` you'll have a proper logger initialized hopefully.

gvso: That sounds like good idea. I moved all the validation to `main` and put them all together. I left the piece that obtains the `--steps` in init since that's more about getting a value from the command, which is what the `init` is doing.

cmd/operator/main.go
```diff
@@ -16,46 +16,123 @@ package main
...
+	if flCLI && flHTTPAddr != "" {
+		log.Fatal("only one of -cli or -http-addr can be used")
+	}
+
+	cfg := config.WithValues(flProject, flRegion, flService, flSteps, flInterval)
+	if !cfg.IsValid(flCLI) {
+		log.Fatalf("invalid config values")
+	}
+
+	logger := log.New()
```
ahmetb, line 106: initialize logger first

pkg/config/config.go
```diff
@@ -18,3 +18,47 @@ type Config struct {
 	Metadata *Metadata `json:"metadata" yaml:"metadata"`
 	Rollout  *Rollout  `json:"rollout" yaml:"rollout"`
 }
+
+// WithValues initializes a configuration with the given values.
+func WithValues(project, region, service string, steps []int64, interval int64) *Config {
+	return &Config{
+		Metadata: &Metadata{
```
ahmetb, line 25: do not call it metadata. think of extensibility in the future.

```
rollouts:
- stategy: {steps: ..., interval : .... }
  targets:
  - service:
      name: x
      project: y
  - service:
      name: x
      project: y

- stategy: {steps: ..., interval : .... }
  targets:
  - serviceSelector:
      foo: bar
```

gvso: I will do this in the next PR

PR #6, opened 2020-07-13
config: Make configuration more extensible

gvso:
This allows to specify different ways of filtering services. It also makes filtering of several services possible.

### DIFF
--- a/cmd/operator/main.go
+++ b/cmd/operator/main.go
@@ -119,18 +119,17 @@ func main() {
 }
 
 func runCLI(logger *logrus.Logger, cfg *config.Config) {
-	serviceConfig := &service.Config{
-		Region:      flRegions[0],
-		Project:     cfg.Targets[0].Project,
-		ServiceName: "hello",
-		Strategy:    cfg.Strategy,
+	serviceInfo := &service.Service{
+		Region:  flRegions[0],
+		Project: cfg.Targets[0].Project,
+		Name:    "hello",
 	}
 
-	client, err := run.NewAPIClient(context.Background(), serviceConfig.Region)
+	client, err := run.NewAPIClient(context.Background(), serviceInfo.Region)
 	if err != nil {
 		logger.Fatalf("could not initilize Cloud Run client: %v", err)
 	}
-	roll := rollout.New(client, serviceConfig).WithLogger(logger)
+	roll := rollout.New(client, serviceInfo, cfg.Strategy).WithLogger(logger)
 
 	for {
 		changed, err := roll.Rollout()
@@ -141,7 +140,7 @@ func runCLI(logger *logrus.Logger, cfg *config.Config) {
 			logger.Info("Rollout process succeeded")
 		}
 
-		interval := time.Duration(serviceConfig.Strategy.Interval)
+		interval := time.Duration(cfg.Strategy.Interval)
 		time.Sleep(interval * time.Second)
 	}
 }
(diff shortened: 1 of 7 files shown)


gvso:
I also wrote the piece that retrieves information about the services targeted in the configuration. I think it'd better to include that in the following PR.

cmd/operator/main.go
```diff
@@ -132,7 +141,39 @@ func runCLI(logger *logrus.Logger, cfg *config.Config) {
 			logger.Info("Rollout process succeeded")
 		}
 
-		interval := time.Duration(cfg.Rollout.Interval)
+		interval := time.Duration(serviceConfig.Strategy.Interval)
 		time.Sleep(interval * time.Second)
 	}
 }
+
+func flagsAreValid() (bool, error) {
```
ahmetb, line 149: might wanna also check flRegions doesn't contain `""` empty strings e.g. `-flag=,,,,`

pkg/service/service.go
```diff
@@ -0,0 +1,20 @@
+package service
+
+import (
+	runapi "github.com/GoogleCloudPlatform/cloud-run-release-operator/internal/run"
+	"github.com/GoogleCloudPlatform/cloud-run-release-operator/pkg/config"
+)
+
+// Config is the configuration for a single service in a specific region.
+type Config struct {
```
ahmetb, line 9: it's kinda weird that there's a config pkg, and also a service pkg that has a `Config` type in it. I don't get this particular type. :) If it's possible to do things without this type, let's do that.

gvso: Yeah, I also wasn't sure about this and initially had a `config.Service` in the `config` package. My intention is that I can pass each of this service configurations for each filtered service to `rollout.Rollout` once I make concurrent rollouts possible.

Let me think about this.

cmd/operator/main.go
gvso: I was assuming that if flRegions is empty, the operator queries all the available regions.

ahmetb: yeah but what if `-regions=a,,,b`

gvso: Oh, I got you. In theory, we could also check if the regions are valid too. Is that something worth doing?

In the follow-up PR, I would add a list of available regions, so we can use that one if `regions` is empty.

pkg/service/service.go
gvso: I decided to name it `service.Service` since it contains information about a single filtered service

cmd/operator/main.go
ahmetb: > Is that something worth doing?
not needed, API already validates that (probably with a 404 or a dns lookup error saying nonexistingregion-run.googleapis.com not found)

pkg/service/service.go
ahmetb: The part I'm trying to figure out is why do we need this struct? We can just pass this service info to `rollout.Rollout`.

And you can simply do filtering on services by working on *runv1.Service type?

gvso: I was probably thinking too ahead of what's needed. When filtering the services, we will get the `run.Service` object, so we probably don't need to query for service in `rollout` again. That's where having this `struct` made it easier to update the code.

It might actually be better to retrieve the `run.Service` again to diminish the possibility of a deployment while running the tool + it simplifies the logic to recover when new deployment occurs during update.

ahmetb: I'd say do not fetch the same service more than once during a single execution.

gvso: Alright. I think I can do both things: remove the struct and fetch service just once. I would just add a way to include more info for the rollout using things like `WithServiceInfo`

ahmetb:
This good to go?

gvso:
I think it is good to go for now

PR #7, opened 2020-07-13
metrics: Obtain error rate and latency

gvso:
This adds functions to retrieve server error rate and request latency. It allows 3 combinations of aligners and reducers that are common for request latency.


internal/metric/wrapper.go
```diff
@@ -0,0 +1,208 @@
...
+
+// Filter is the filter for the query.
+type Filter struct {
+	service  string
+	revision string
+	query    string
+}
+
+// Metric types.
+const (
```
ahmetb, line 35: might be wise to call this package `stackdriver`, and have `metrics` package just be the generic interfaces (so you can replace stackdriver with mocks etc).

internal/metric/wrapper.go
```diff
@@ -0,0 +1,208 @@
...
+)
+
+// Client represents a wrapper around the Cloud Monitoring package.
+type Client interface {
+	Latency(ctx context.Context, filter Filter, startTime time.Time) (float64, error)
+}
+
+// API is a wrapper for the Cloud Monitoring package.
+type API struct {
+	Client  *monitoring.MetricClient
```
ahmetb, line 23: you can embed this probably

internal/metric/wrapper.go
```diff
@@ -0,0 +1,208 @@
...
+func NewFilter(serviceName, revisionName string) Filter {
+	filter := Filter{}
+	filter = filter.Add("resource.labels.service_name", serviceName)
+	filter = filter.Add("resource.labels.revision_name", revisionName)
+
+	return filter
+}
+
+// Add adds a filter for the query.
+func (f Filter) Add(key string, value string) Filter {
```
ahmetb, line 201: might be wiser to call the type `Query` and this method `Filter`. e.g.

NewQuery(svc).Filter(k,v).Filter(k2,v2)

internal/metric/wrapper.go
```diff
@@ -0,0 +1,208 @@
...
+	if f.query != "" {
+		f.query += " AND "
+	}
+	f.query += fmt.Sprintf("%s=%q", key, value)
```
ahmetb, line 205: for now we probably just do `%s AND %s`, but in the future, this could benefit from an internal variable storing the filters, and a `Query() string` method compiling the internal state into a query string.

internal/metric/wrapper.go
```diff
@@ -0,0 +1,208 @@
...
+const (
+	RequestLatencies = "run.googleapis.com/request_latencies"
+	RequestCount     = "run.googleapis.com/request_count"
+)
+
+type alignReduce int32
+
+// Series aligner and cross series reducer types (for latency).
+const (
+	Align99Reduce99 alignReduce = 1
```
ahmetb, line 44: I'm not aware of aligning and reducing terminology. What's the difference?

gvso: When obtaining metrics, you set an interval for the time you're interested in (in the Cloud Run Metrics graph, for instance, you have 1 day, 1 hour, etc.). Of course, this interval is customizable.

The interval can be broken up in pieces determined by `alignmentPeriod`. For instance, you can break an interval of 1h in pieces of 20 minutes each. Doing so you get a total of `interval / alignmentPeriod` points. Each point represents what happened during `alignmentPeriod` seconds. Of course, if you set `alignmentPeriod=1200s` (20 minutes), there might be a lot of requests during those 20 minutes. How do you get one single point for entire 20-minute period? By using the per series **aligner**. The aligner is basically the function that determines how to aggregate all those requests in the 20-minute period and pick one value.

But you get a total of `interval / alignmentPeriod` points. That's where the cross series reducer helps you get one single point based on a grouping fields like response code class (2xx, 5xx, etc.).

internal/metric/wrapper.go
gvso: I agree. I didn't want to "overengineer" here since we just need simple queries for now.

internal/metrics/metrics.go
```diff
@@ -0,0 +1,29 @@
...
+const (
+	Align99Reduce99 AlignReduce = 1
+	Align95Reduce95             = 2
+	Align50Reduce50             = 3
+)
+
+// Client represents a wrapper around the Cloud Monitoring package.
+type Client interface {
+	Latency(ctx context.Context, query Query, startTime time.Time, alignReduceType AlignReduce) (float64, error)
+	ServerErrorRate(ctx context.Context, query Query, startTime time.Time) (float64, error)
```
ahmetb, line 28: ErrorRate + add some godoc on behavior (non-2xx)

internal/stackdriver/stackdriver.go
```diff
@@ -0,0 +1,196 @@
...
+
+// Query is the filter used to retrieve metrics data.
+type Query struct {
+	filter string
+}
+
+// Metric types.
+const (
+	RequestLatencies = "run.googleapis.com/request_latencies"
+	RequestCount     = "run.googleapis.com/request_count"
```
ahmetb, line 31: unexport

internal/metrics/metrics.go
```diff
@@ -0,0 +1,29 @@
...
+// Series aligner and cross series reducer types (for latency).
+const (
+	Align99Reduce99 AlignReduce = 1
+	Align95Reduce95             = 2
+	Align50Reduce50             = 3
+)
+
+// Client represents a wrapper around the Cloud Monitoring package.
+type Client interface {
+	Latency(ctx context.Context, query Query, startTime time.Time, alignReduceType AlignReduce) (float64, error)
```
ahmetb, line 27: accept `offset time.Duration`

internal/metrics/metrics.go
```diff
@@ -0,0 +1,29 @@
+package metrics
+
+import (
+	"context"
+	"time"
+)
+
+// Query is the filter used to retrieve metrics data.
+type Query interface {
+	Filter(key, value string)
```
ahmetb, line 10: return Query from this instead of modifying internally.

internal/stackdriver/stackdriver.go
```diff
@@ -0,0 +1,196 @@
...
+func NewQuery(serviceName, revisionName string) Query {
+	query := Query{}
+	query.Filter("resource.labels.service_name", serviceName)
+	query.Filter("resource.labels.revision_name", revisionName)
+
+	return query
+}
+
+// Filter adds a filter for the query.
+func (q *Query) Filter(key, value string) {
```
ahmetb, line 186: make it immutable by removing the ptr receiver.

internal/stackdriver/stackdriver.go
```diff
@@ -0,0 +1,196 @@
...
+	}
+
+	return
+}
+
+// NewQuery initializes a query.
+func NewQuery(serviceName, revisionName string) Query {
+	query := Query{}
+	query.Filter("resource.labels.service_name", serviceName)
+	query.Filter("resource.labels.revision_name", revisionName)
```
ahmetb, line 180: add project_id filter.

internal/stackdriver/stackdriver.go
```diff
@@ -0,0 +1,196 @@
...
+		if err != nil {
+			return 0, errors.Wrap(err, "error when iterating through time series")
+		}
+
+		// Because the interval and the series aligner are the same, only one
+		// point is returned per time series.
+		switch series.Metric.Labels["response_code_class"] {
+		case "2xx":
+			successfulResponseCount += series.Points[0].Value.GetInt64Value()
+			break
```
ahmetb, line 140: default:

gvso:
Ignore this issue for now. I'll come back to it later and use MQL directly

ahmetb:
cc: @grayside this is mostly reviewed so feel free to take a look.
we created an interface for metrics reading w/ stackdriver impl.

internal/metrics/metrics.go
gvso: I have thought about this, and I believe it makes more sense from the user of this interface to pass a startTime (`time.Time`) value since that would likely be the last time the metrics were checked. 

If the implementation of, say prometheus, uses duration, the implementation should take care of converting the startTime value to a `time.Duration`.

ahmetb: > I have thought about this, and I believe it makes more sense from the user of this interface to pass a startTime (time.Time) value since that would likely be the last time the metrics were checked.

last time I recall our design was that the operator checks last N minutes; not "since last time checked". Storing such info probably requires additional storage? I also don't see a downside of looking at last N minute window.

gvso: AFAIK, we didn't discuss exactly how to estimate the time interval between metrics checks. Both ways should work fine since I think, probabilistically, there might not be any difference in the results for slightly different times. The time intervals might even be similar in many (most) cases.

The easiest/simpler one to implement for now is to check last N minutes though. Let's do that.

PR #8, opened 2020-07-14
service: Filter several services using config

gvso:
This uses the targets configuration and concurrently filters the services based on the selectors.


internal/run/wrapper.go
```diff
@@ -21,6 +21,22 @@ type API struct {
 	Region string
 }
 
+// Regions are the available regions.
+var Regions = []string{
```
ahmetb, line 25: this can be fetched from api.

cmd/operator/main.go
```diff
@@ -119,19 +117,10 @@ func main() {
...
-		Region:      flRegions[0],
-	}, cfg.Strategy).WithLogger(logger)
-
 	for {
+		services := service.Filter(cfg)
+		roll := rollout.New(services[0], cfg.Strategy).WithLogger(logger)
+
 		changed, err := roll.Rollout()
 		if err != nil {
 			logger.Infof("Rollout failed: %v", err)
```
ahmetb, line 126: re: this code block (you can address in a different PR)

```
		interval := time.Duration(cfg.Strategy.Interval)
		time.Sleep(interval * time.Second)

```

we need to find a better name for "Interval". (e.g. "What interval is this for?")

internal/run/wrapper.go
```diff
@@ -47,6 +63,12 @@ func (a *API) ReplaceService(namespace, serviceID string, svc *run.Service) (*ru
 	return a.Client.Namespaces.Services.ReplaceService(serviceName, svc).Do()
 }
 
+// ListServices gets services filtered by a label.
+func (a *API) ListServices(namespace string, labelSelector string) (*run.ListServicesResponse, error) {
+	parent := fmt.Sprintf("namespaces/%s", namespace)
+	return a.Client.Namespaces.Services.List(parent).LabelSelector(labelSelector).Do()
```
ahmetb, line 69: interesting that we support server-side label selector!

cmd/operator/main.go
gvso: Yes, we should probably make it clearer. We can change it in another PR since this interval is only used here.

ahmetb, line 131: This probably should be fatalf.

cmd/operator/main.go
```diff
@@ -119,19 +117,15 @@ func main() {
...
-		RunClient:   runclient,
-		Project:     cfg.Targets[0].Project,
-		ServiceName: "hello",
-		Region:      flRegions[0],
-	}, cfg.Strategy).WithLogger(logger)
-
 	for {
+		services := service.Filter(logger, cfg)
+		if len(services) == 0 {
+			logger.Info("No service matches the targets")
```
ahmetb, line 123: this is probably a warn

cmd/operator/main.go
```diff
@@ -119,19 +117,15 @@ func main() {
...
+			interval := time.Duration(cfg.Strategy.Interval)
```
ahmetb, line 124: try not to repeat code. this part is repeated right now

ahmetb:
cc: @grayside if you have time to take a look. this PR is about fetching and filtering Cloud Run services from the API.

pkg/service/service.go
```diff
@@ -1,13 +1,125 @@
...
+		RunClient:   runclient,
+		Service:     svc,
+		Project:     project,
+		ServiceName: serviceName,
+		Region:      region,
+	}
+}
+
+// Filter retrieves all the services that match the filter in the configuration.
+func Filter(logger *logrus.Logger, cfg *config.Config) []*Client {
```
ahmetb, line 35: 1. I think this method is poorly named. It sounds like QueryAndFilterAllServices()

2. You only use `cfg.Targets` so why are you passing `config.Config` here?

pkg/service/service.go
```diff
@@ -1,13 +1,125 @@
...
+	clientsCh := make(chan []*Client)
```
ahmetb, line 36: this code is hard to follow wit channels, I'm not sure they are required here. what are you trying to do? 

if all you're trying to do is get services in parallel, you should use WaitGroup + Goroutines; not channel.

pkg/service/service.go
```diff
@@ -1,13 +1,125 @@
...
+	numberOfFilterByRegionCalls := 0
```
ahmetb, line 37: `filterByRegionCall` identifier also doesn't make to me, isn't the task here simply:

1. Identify regions (not this method)
2. Choose which regions to use if a region list is specified (not this method)
3. Query all services in all regions "in parallel" + wait
4. As part of (3), query services in 1 region, with label filter.

so I'm not sure which one of 1,2,3,4 I'm reading right now. It seems it's sort of all. :)

pkg/service/service.go
```diff
@@ -1,13 +1,125 @@
...
+		"project":        project,
+		"region":         region,
+		"selectorType":   selector.Type,
+		"selectorFilter": selector.Filter,
+	})
+
+	runclient, err := runapi.NewAPIClient(context.Background(), region)
+	if err != nil {
+		lg.Errorf("can not initialize Cloud Run client: %v", err)
+		clientsCh <- nil
```
ahmetb, line 83: don't use channels, they aren't required.

pkg/service/service.go
```diff
@@ -1,13 +1,125 @@
...
+	}
+
+	close(clientsCh)
+
+	return svcClients
+}
+
+// filterByRegion searches for a service name or label selector in an specific
+// regional endpoint.
+func filterByRegion(clientsCh chan []*Client, logger *logrus.Logger, project, region string, selector config.TargetSelector) {
```
ahmetb, line 72: again this name isn't good. I can't understand what this method by looking at the implementation.

I see "service name type" but we said we'll do label based selectors. I'm not entirely sure what's happening here. Furthermore `chan`s just contribute to the complexity.

ahmetb:
Right now, I'm seeing overcomplicated code in this PR getting close to spaghetti. :) It would be good to go to the drawing board and rethink some of these packages and split them into smaller methods.

Now that the algorithms are getting hard to follow, it usually points to _code smell_. What we are trying to do here is fairly simple: Query all services in some/all regions with a label selector. That's it. But the code is far beyond that.

pkg/service/service.go
```diff
@@ -1,13 +1,125 @@
...
-// Client is the connection to update and obtain data about a single service.
+// Client is the connection to update and obtain data about a single service. It
+// contains a client to the Cloud Run API, information about the service and the
+// service object.
 type Client struct {
 	RunClient   runapi.Client
+	Service     *run.Service
 	Project     string
 	ServiceName string
 	Region      string
```
grayside, line 20: Given project, name, and region are all things that can be found on the service object, is there a value in duplicate data? If not, maybe these would be better served by helper methods.

pkg/service/service.go
gvso: The major reason this code looks confusing is that I'm using channels to retrieve the service objects I am getting from the API. If I run this in goroutines, what's the best way to store the returned values from the API using a WaitGroup? Using global state? Creating a struct with a `[]run.Service/[]service.Client` field + sync.WaitGroup field? Passing the array storing the values and using mutex?

gvso:
@ahmetb The major reason this code looks confusing is that I'm using channels to retrieve the service objects I am getting from the API. If I run this in goroutines, what's the best way to store the returned values from the API using a WaitGroup? Using global state? Creating a struct with a []run.Service/[]service.Client field + sync.WaitGroup field? Passing the array storing the values and using mutex?

The sources I have seen either don't have my use case/use the channel implementation. I'd appreciate any guidelines here since I haven't done concurrency in real-world applications (only at school labs).

cmd/operator/main.go
```diff
@@ -119,19 +117,15 @@ func main() {
...
+			time.Sleep(interval * time.Second)
+		}
+		roll := rollout.New(services[0], cfg.Strategy).WithLogger(logger)
```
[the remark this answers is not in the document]
gvso: This is definitively a TODO. I didn't want to overcomplicate this PR since rollout of several services would be in another PR. I'll add a TODO here. Thanks!

pkg/service/service.go
```diff
@@ -1,13 +1,125 @@
...
+	"context"
+
 	runapi "github.com/GoogleCloudPlatform/cloud-run-release-operator/internal/run"
+	"github.com/GoogleCloudPlatform/cloud-run-release-operator/pkg/config"
+	"github.com/sirupsen/logrus"
+	"google.golang.org/api/run/v1"
 )
 
-// Client is the connection to update and obtain data about a single service.
+// Client is the connection to update and obtain data about a single service. It
```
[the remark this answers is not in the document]
gvso: That's possible. However, just take into consideration that we probably want to run the rollout process concurrently, so we need a `service.Client` per filtered service.

That being said, I would probably need to create a cache of `run.Client`s for regions the filter has seen already. Is it worth doing so in this PR or should we do it in another one?

pkg/service/service.go
gvso: I'm not sure that all these data can be extracted from the service object. Looking at the YAML in the Cloud Run service, I get

```yaml
apiVersion: serving.knative.dev/v1
kind: Service
metadata:
  name: hello
  namespace: '929898821634'
  generation: 130
  creationTimestamp: '2020-06-29T22:14:02.259763Z'
  labels:
    rollout-strategy: gradual
    cloud.googleapis.com/location: us-east1
  annotations:
    run.googleapis.com/client-name: gcloud
    rollout.cloud.run/stableRevision: hello-00124-cim
    client.knative.dev/user-image: gcr.io/cloudrun/hello
    run.googleapis.com/client-version: 298.0.0
    run.googleapis.com/launch-stage: ALPHA
```
Can we trust that all types of deployment (not using gcloud, for instance) are going to include the `name` and `namespace` metadata? The region looks like a label that was created by my deployment tool (gcloud).

ahmetb:
The only variable in these parallel executions is region name. So you don't need an array of client instances, you can init them in Goroutines. Would be good to study some sync.WorkGroup examples on the internet. You can simply append the resulting svc records into an array.

pkg/service/service.go
ahmetb: name/namespace are mandatory.
however don't rely on Google Cloud specific annotations since we might support knative someday.

If you keep passing `<region,name,project>` triplet around, just create a `ServiceRecord` struct?

pkg/service/service.go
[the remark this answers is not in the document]
gvso: After discussing it with Ahmet, I think there's no harm in creating as many `run.Client`s as needed since initializing them does not make any HTTP requests per sé. So, there's no connection that should be closed or anything like that, and it's not using much memory. Therefore, I think it's better to creating the object as needed to increase code readability.

cmd/operator/service.go
```diff
@@ -0,0 +1,92 @@
+package main
```
gvso, line 1: Should I keep this file or just move the content to the `main.go` file?

ahmetb: We can move. Doesn't make a lot of difference for now.

cmd/operator/service.go
```diff
@@ -0,0 +1,94 @@
...
+	}
+
+	var records []*rollout.ServiceRecord
+	for _, svc := range svcs {
+		records = append(records, &rollout.ServiceRecord{
+			Service: svc,
+			Project: project,
+			Region:  region,
+		})
+	}
```
ahmetb, line 78: you prob need a ToServiceRecord method to call from goroutine, this doesn't have to be here.

cmd/operator/service.go
```diff
@@ -0,0 +1,94 @@
...
+
+	return records, nil
+}
+
+func getRegions(target *config.Target) ([]string, error) {
```
ahmetb, line 83: add comment

cmd/operator/service.go
```diff
@@ -0,0 +1,94 @@
...
+	regions := target.Regions
+	if len(regions) == 0 {
```
ahmetb, line 85: flip this !=

internal/run/wrapper.go
```diff
@@ -47,6 +50,44 @@ func (a *API) ReplaceService(namespace, serviceID string, svc *run.Service) (*ru
...
+	var services []*run.Service
+	for _, service := range servicesList.Items {
+		services = append(services, service)
+	}
+	return services, nil
+}
+
+// Regions gets the supported regions for the project.
+func Regions(project string) ([]string, error) {
+	if len(regions) == 0 {
```
ahmetb, line 71: flip !=

ahmetb: add logs about using cached regions

internal/util/types.go
```diff
@@ -0,0 +1,8 @@
+package util
+
+type contextKey string
+
+// Context keys.
+const (
```
gvso, line 6: This is used in only one place as of now, but it'll be useful when adding the health package and logs in metrics/stackdriver

cmd/operator/service.go
```diff
@@ -83,18 +88,22 @@ func getServicesByRegionAndLabel(ctx context.Context, logger *logrus.Logger, pro
 // If the target configuration does not specify any regions, the entire list of
 // regions is retrieved from API.
 func determineRegions(ctx context.Context, logger *logrus.Logger, target *config.Target) ([]string, error) {
+	lg := logger.WithField("labelSelector", target.LabelSelector)
```
ahmetb, line 91: label selector is not applicable to determining regions. so probably this log stmt shouldn't have this field

cmd/operator/main.go
```diff
@@ -119,32 +110,40 @@ func main() {
...
 			logger.Info("Rollout process succeeded")
 		}
 
-		interval := time.Duration(cfg.Strategy.Interval)
-		time.Sleep(interval * time.Second)
+		sleepInSeconds(cfg.Strategy.Interval)
 	}
 }
 
+func sleepInSeconds(interval int64) {
```
ahmetb, line 142: no need to create this method. this is already 1-liner

cmd/operator/service.go
```diff
@@ -0,0 +1,117 @@
...
+		retServices []*rollout.ServiceRecord
+		retError    error
+		mu          sync.Mutex
+		wg          sync.WaitGroup
+	)
+
+	for _, target := range targets {
+		regions, err := determineRegions(ctx, logger, target)
+		if err != nil {
+			logger.Errorf("cannot determine regions: %v", err)
```
ahmetb, line 33: this is probably a `return errors.Wrap`.

fail early fail often.

cmd/operator/service.go
```diff
@@ -0,0 +1,117 @@
...
+
+// getServicesByRegionAndLabel returns all the service records that match the
+// labelSelector in a specific region.
+func getServicesByRegionAndLabel(ctx context.Context, logger *logrus.Logger, project, region, labelSelector string) ([]*run.Service, error) {
+	lg := logger.WithFields(logrus.Fields{
+		"region":        region,
+		"labelSelector": labelSelector,
+	})
+
+	lg.Debugf("filtering services")
```
ahmetb, line 71: this is incorrect because you don't have any services to filter yet. :)

ahmetb: `querying Cloud Run services` would be an appropriate statement

cmd/operator/service.go
```diff
@@ -0,0 +1,117 @@
...
+
+// determineRegions gets the regions the label selector should be searched at.
+//
+// If the target configuration does not specify any regions, the entire list of
+// regions is retrieved from API.
+func determineRegions(ctx context.Context, logger *logrus.Logger, target *config.Target) ([]string, error) {
+	lg := logger.WithField("labelSelector", target.LabelSelector)
+	regions := target.Regions
+	if len(regions) != 0 {
+		lg.Debug("regions set in configuration")
```
ahmetb, line 94: `using predefined list of regions, skip querying from API` is a more descriptive

internal/util/types.go
```diff
@@ -0,0 +1,8 @@
...
+	ContextKeyLogger = contextKey("logger")
```
ahmetb, line 7: this pkg is nor useful right now.

skip for now. create new PR where you properly design a `util.WithLogger(ctx, logger)` method.

ContextKeyLogger is an implementation detail that should not be exposed to the outside world.

cmd/operator/main.go
```diff
@@ -119,32 +110,40 @@ func main() {
 }
 
 func runCLI(logger *logrus.Logger, cfg *config.Config) {
```
ahmetb, line 112: this probably can return an `error`. use `log.fatal`s sparingly because they ignore all call stack and finish the process immediately.

if you use `return`s as much as we can, we can do `log.fatal` in max 1-2 places, and have a chance to fail with either just the error message and potentially also with the callstack in debug mode.

internal/run/wrapper.go
```diff
@@ -21,6 +21,9 @@ type API struct {
 	Region string
 }
 
+// regions are the available regions.
+var regions = []string{}
```
ahmetb, line 25: can you add a `// TODO caching regions might be unnecessary if we are querying them once during the lifespan of the process`

internal/run/wrapper.go
```diff
@@ -47,6 +50,41 @@ func (a *API) ReplaceService(namespace, serviceID string, svc *run.Service) (*ru
...
+// ServicesWithLabelSelector gets services filtered by a label selector.
+func (a *API) ServicesWithLabelSelector(namespace string, labelSelector string) ([]*run.Service, error) {
+	parent := fmt.Sprintf("namespaces/%s", namespace)
+
+	servicesList, err := a.Client.Namespaces.Services.List(parent).LabelSelector(labelSelector).Do()
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to filter services by label selector")
+	}
+
+	return servicesList.Items, nil
```
ahmetb, line 62: this is something to learn, you can do:

`return servicesList.Items, errors.Wrap(err, "..")`

Wrap returns nil if err is nil. But make sure `servicesList` isn't a pointer type that can panic in case of error.

internal/run/wrapper.go
```diff
@@ -47,6 +50,41 @@ func (a *API) ReplaceService(namespace, serviceID string, svc *run.Service) (*ru
...
+}
+
+// Regions gets the supported regions for the project.
+func Regions(project string) ([]string, error) {
+	if len(regions) != 0 {
+		return regions, nil
+	}
+
+	client, err := run.NewService(context.Background())
```
ahmetb, line 71: avoid `context.Background` pass `ctx` here. That's why we have ctx, if we don't use it, it's pointless.

internal/run/wrapper.go
gvso: That's helpful and I didn't know about it. I wonder if in this case it is better as it is since `servicesList` is a pointer, so I will have add an if anyways. Since is more common to see check for errors, I think an `if err != nil` make it more readable.

ahmetb: If servicesList.Items has a chance of panicking, don't do what I said. :) it works otherwise however

cmd/operator/main.go
gvso: That's a good point. I'll change it once we start handling several services. The `log.Fatal` when rollout fails will go away since rollout of service A shouldn't probably affect rollout of service B. 

But yeah, there are more cases where we can error entirely (e.g. fail to filter targeted services)

cmd/operator/service.go
```diff
@@ -76,13 +76,13 @@ func getServicesByRegionAndLabel(ctx context.Context, project, region, labelSele
 //
 // If the target configuration does not specify any regions, the entire list of
```
ahmetb, line 77: fix in a separate PR please:
this is not proper godoc. godocs start with the method's name.

PR #9, opened 2020-07-17
rollout: Log events in the rollout process

gvso:
This is an attempt to start logging more information about the events occurring during the rollout process. It also adds the possibility to adjust the logging level through a CLI flag

### DIFF
--- a/cmd/operator/main.go
+++ b/cmd/operator/main.go
@@ -53,10 +53,17 @@ func (steps stepFlags) String() string {
 }
 
 var (
+	logrusLevel    logrus.Level
+	flLoggingLevel string
+
 	flCLI           bool
 	flHTTPAddr      string
 	flProject       string
 	flLabelSelector string
+
+	// Either service or label selection needed.
+	flService string
+	flLabel   string
 
 	// Empty array means all regions.
 	flRegions       []string
@@ -69,6 +76,7 @@ var (
 )
 
 func init() {
+	flag.StringVar(&flLoggingLevel, "verbosity", "info", "the logging level (e.g. debug)")
 	flag.BoolVar(&flCLI, "cli", false, "run as CLI application to manage rollout in intervals")
 	flag.StringVar(&flHTTPAddr, "http-addr", "", "listen on http portrun on request (e.g. :8080)")
 	flag.StringVar(&flProject, "project", "", "project in which the service is deployed")
@@ -90,6 +98,8 @@ func main() {
 	if !valid {
 		logger.Fatalf("invalid flags: %v", err)
 	}
+
+	logger.SetLevel(logrusLevel)
 
 	// Configuration.
 	target := config.NewTarget(flProject, flRegions, flLabelSelector)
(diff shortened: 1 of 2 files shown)


cmd/operator/main.go
```diff
@@ -160,6 +166,12 @@ func flagsAreValid() (bool, error) {
 		}
 	}
 
+	var err error
+	logrusLevel, err = logrus.ParseLevel(flLoggingLevel)
+	if err != nil {
+		return false, errors.Wrap(err, "invalid verbosity value")
```
ahmetb, line 172: this should be art of init/main CLI flag parsing. as you can imagine, it's too late to initialize a logger here.

you initialize the logger in main() so why are you doing it here?

pkg/rollout/rollout.go
```diff
@@ -212,6 +221,7 @@ func (r *Rollout) nextCandidateTraffic(current int64) int64 {
 
 // updateAnnotations updates the annotations to keep some state about the rollout.
 func (r *Rollout) updateAnnotations(svc *run.Service, stable, candidate string) *run.Service {
+	r.Log.Debug("updating annotations for the service")
```
ahmetb, line 224: this sounds like you're making an api call

pkg/rollout/rollout.go
```diff
@@ -137,10 +146,10 @@ func (r *Rollout) SplitTraffic(svc *run.Service, stable, candidate string) *run.
 	traffic = append(traffic, &run.TrafficTarget{LatestRevision: true, Tag: LatestTag})
 
 	if !r.promoteToStable {
-		r.Log.Infof("Assigning %d%% of the traffic to stable revision %s", stablePercent, stable)
-		r.Log.Infof("Assigning %d%% of the traffic to candidate revision %s", candidateTraffic.Percent, candidate)
+		r.Log.Infof("assigning %d%% of the traffic to stable revision %s", stablePercent, stable)
+		r.Log.Infof("assigning %d%% of the traffic to candidate revision %s", candidateTraffic.Percent, candidate)
 	} else {
-		r.Log.Infof("Making revision %s stable", candidate)
+		r.Log.Infof("making revision %s stable", candidate)
```
ahmetb, line 152: "will make" is more appropriate, bc you aren't making api call right now

cmd/operator/main.go
gvso: It could go in `main`. The only small side consequence is that it'd add an extra `Fatal` call.

ahmetb: that's fine, those are early in the program, so we'd much rather fail early before we set up other things.

pkg/rollout/rollout.go
gvso: I also have that feeling. Is the word "updating" or the verb tense for you? For me is the word "updating"

What about "including annotations in the service"?

ahmetb: I wouldn’t add logs to this method. It does the same thing always. Doesn't have different code paths. We always call this.

PR #11, opened 2020-07-17
mock: Create an internal mock package

gvso:
This creates a new internal package for mocks. This will be more useful as we add mocks for the metrics and health package (maybe others) in the future.

### DIFF
--- a/internal/mock/run.go
+++ b/internal/mock/run.go
@@ -2,8 +2,8 @@ package mock
 
 import "google.golang.org/api/run/v1"
 
-// RunAPI represents a mock implementation of run.API.
-type RunAPI struct {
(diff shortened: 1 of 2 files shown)


ahmetb:
@Grayside do u know it's more idiomatic to have mocks in each package (or subpkg) as opposed to in a centralized package? This doesn't feel right.

grayside:
A central mock seems less idiomatic to me as well: if we're mocking it centrally that implies our testing is shallow or the the design has poor modularity. Instead of passing the full RunAPI struct around, maybe there should be ServiceUpdater and TrafficSplitter interfaces.

gvso:
Thank you both! I'm closing this PR since it is not appropriate.

PR #12, opened 2020-07-20
config: Add metrics checks in configuration

gvso:
This allows to specify metrics checks in the configuration. This will be used by the health package to determine if the candidate meets the metrics checks criteria

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -1,12 +1,12 @@
 package config
 
-// MetricsCheck is metrics check type.
+// MetricsCheck is the metrics check type.
 type MetricsCheck string
 
 // Supported metrics checks.
 const (
-	LatencyCheckType MetricsCheck = "request-latency"
-	ErrorCheckType                = "error-rate"
+	LatencyMetricsCheck MetricsCheck = "request-latency"
+	ErrorMetricsCheck                = "error-rate"
 )
 
 // Target is the configuration to filter services.
@@ -24,18 +24,19 @@ type Target struct {
 	LabelSelector string
 }
 
-// Check is a metrics threshold that should be met to consider a candidate
+// Metric is a metrics threshold that should be met to consider a candidate
 // healthy.
-type Check struct {
+type Metric struct {
 	Type       MetricsCheck
 	Percentile float64
+	Min        float64
 	Max        float64
 }
 
 // Strategy is the steps and configuration for rollout.
 type Strategy struct {
-	Steps  []int64
-	Checks []Check
+	Steps   []int64
+	Metrics []Metric
 
 	// TODO: Give this property a clearer name.
 	Interval int64
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -1,5 +1,14 @@
 package config
 
+// MetricsCheck is metrics check type.
+type MetricsCheck string
+
+// Supported metrics checks.
+const (
+	LatencyCheckType MetricsCheck = "request-latency"
+	ErrorCheckType                = "error-rate"
+)
+
 // Target is the configuration to filter services.
 //
 // A target might have the following form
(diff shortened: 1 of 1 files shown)


pkg/config/config.go
```diff
@@ -15,9 +24,18 @@ type Target struct {
 	LabelSelector string
 }
 
+// Check is a metrics threshold that should be met to consider a candidate
+// healthy.
+type Check struct {
+	Type       MetricsCheck
+	Percentile float64
+	Max        float64
```
ahmetb, line 32: What does max do

pkg/config/config.go
```diff
@@ -15,9 +24,18 @@ type Target struct {
...
+}
+
 // Strategy is the steps and configuration for rollout.
 type Strategy struct {
-	Steps []int64
+	Steps  []int64
+	Checks []Check
```
ahmetb, line 38: We can probably find a better name. There's health and monitoring involved but the name doesn't reflect.

pkg/config/config.go
gvso: I was following what we currently have in the design doc.

```
strategy:
   ...
   checks:
       - type: request-latency
         percentile: 99
         max: 500
       - type: error-rate
         max: 1
```

Max is the maximum allowed value for each of the health checks. For example, I want the maximum allowed error rate to be 1%.

pkg/config/config.go
gvso: `health-checks` or `metrics-checks`? Flagger is using `metrics` in their [config definition](https://docs.flagger.app/usage/how-it-works).

pkg/config/config.go
ahmetb: should prob call `error-rate-percent` to make it clear

also how do I represent "min N requests served" in this model?

pkg/config/config.go
ahmetb: `metrics` or `health` would be good IMO

pkg/config/config.go
gvso: Min is still not supported in this model. I would include it in the PR for request count.

Also, I think the request count should be something that is not a health/metrics check type. For a health check, if the specified condition is not met, a rollback should occur. That's not what we want with the request count. We want to consider the health after we confirmed that the number of requests is what we expect.

If we want to represent the option in YAML format (not a thing we are doing now), I'm thinking of two options.
1)
```
strategy:
   ...
   metrics:
       minRequestCount: 1000
       checks (maybe thresholds?):
           - type: request-latency
             percentile: 99
             max: 500
```

2)

```
strategy:
   steps: [5, 30, 60]
   metricsMinRequestCount: 1000
   metrics:
       - type: request-latency
         percentile: 99
         max: 500
       - type: error-rate
         max: 1
```

ahmetb: technically, it's a check. And you have checks[]. why is this not an option:

```
- type: request-count
  min: 1000
```

gvso: I could do that, but it could be confusing since everything you specified in `metrics` are things that, if not met, should produce a rollback. In the case of the request count, we don't want a rollback because your revision didn't get 1000 requests.

But doing it this way or the ways I showed before is only a concern when dealing with YAML configurations. We could handle that later.

I'll ad a Min field in the metrics check struct since we might add some metrics checks that need it in the future anyways (if request count is not one of them)

ahmetb: Valid point. We should've thought this as part of the "design". :) It might be go back to the drawing board at this stage.

ahmetb: I'll let this one pass for now. 

We should add probably `minRequestCount` as a check in the future as well, but it would have an implicit "triggersRollback: false" like feature.

PR #13, opened 2020-07-21
metrics: Obtain request count

gvso:
This adds a function to the Metrics interface and adds an implementation in stackdriver to obtain the number of requests for a given offset.

### DIFF
--- a/internal/metrics/metrics.go
+++ b/internal/metrics/metrics.go
@@ -24,18 +24,17 @@ const (
 )
 
 // Metrics represents a monitoring API such as Stackdriver.
-//
-// RequestCount count returns the number of requests for the given offset and
-// query.
-//
-// Latency returns the request latency after applying the specified series
-// aligner and cross series reducer. The result is in milliseconds.
-//
-// Error rate gets all the server responses. It calculates the error rate by
-// performing the operation (5xx responses / all responses).
 type Metrics interface {
+	// RequestCount count returns the number of requests for the given offset
+	// and query.
 	RequestCount(ctx context.Context, query Query, offset time.Duration) (int64, error)
+
+	// Latency returns the request latency after applying the specified series
+	// aligner and cross series reducer. The result is in milliseconds.
 	Latency(ctx context.Context, query Query, offset time.Duration, alignReduceType AlignReduce) (float64, error)
+
+	// Error rate gets all the server responses. It calculates the error rate by
+	// performing the operation (5xx responses / all responses).
 	ErrorRate(ctx context.Context, query Query, offset time.Duration) (float64, error)
 }
 
--- a/internal/metrics/metrics.go
+++ b/internal/metrics/metrics.go
@@ -25,12 +25,16 @@ const (
 
 // Metrics represents a monitoring API such as Stackdriver.
 //
+// RequestCount count returns the number of requests for the given offset and
+// query.
+//
 // Latency returns the request latency after applying the specified series
 // aligner and cross series reducer. The result is in milliseconds.
 //
 // Error rate gets all the server responses. It calculates the error rate by
 // performing the operation (5xx responses / all responses).
 type Metrics interface {
+	RequestCount(ctx context.Context, query Query, offset time.Duration) (int64, error)
 	Latency(ctx context.Context, query Query, offset time.Duration, alignReduceType AlignReduce) (float64, error)
 	ErrorRate(ctx context.Context, query Query, offset time.Duration) (float64, error)
 }
--- a/internal/metrics/metrics.go
+++ b/internal/metrics/metrics.go
@@ -25,16 +25,15 @@ const (
 
 // Metrics represents a monitoring API such as Stackdriver.
 type Metrics interface {
-	// RequestCount count returns the number of requests for the given offset
-	// and query.
+	// Returns the number of requests for the given offset and query.
 	RequestCount(ctx context.Context, query Query, offset time.Duration) (int64, error)
 
-	// Latency returns the request latency after applying the specified series
-	// aligner and cross series reducer. The result is in milliseconds.
+	// Returns the request latency after applying the specified series aligner
+	// and cross series reducer. The result is in milliseconds.
 	Latency(ctx context.Context, query Query, offset time.Duration, alignReduceType AlignReduce) (float64, error)
 
-	// Error rate gets all the server responses. It calculates the error rate by
-	// performing the operation (5xx responses / all responses).
+	// Gets all the server responses and calculates the error rate by performing
+	// the operation (5xx responses / all responses).
 	ErrorRate(ctx context.Context, query Query, offset time.Duration) (float64, error)
 }
 
--- a/internal/stackdriver/stackdriver.go
+++ b/internal/stackdriver/stackdriver.go
@@ -8,6 +8,7 @@ import (
 	"github.com/GoogleCloudPlatform/cloud-run-release-operator/internal/metrics"
 	"github.com/GoogleCloudPlatform/cloud-run-release-operator/internal/util"
 	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
 
 	// TODO: Migrate to cloud.google.com/go/monitoring/apiv3/v2 once RPC for MQL
 	// query is added (https://cloud.google.com/monitoring/api/ref_v3/rest/v3/projects.timeSeries/query).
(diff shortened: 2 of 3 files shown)


internal/metrics/metrics.go
```diff
@@ -23,12 +23,16 @@ const (
...
 // Metrics represents a monitoring API such as Stackdriver.
 //
+// RequestCount count returns the number of requests for the given offset and
+// query.
+//
 // Latency returns the request latency after applying the specified series
 // aligner and cross series reducer. The result is in milliseconds.
 //
 // Error rate gets all the server responses. It calculates the error rate by
 // performing the operation (5xx responses / all responses).
```
ahmetb, line 33: need to be on the method

internal/stackdriver/stackdriver.go
```diff
@@ -44,6 +44,46 @@ func NewAPIClient(ctx context.Context, project string) (*API, error) {
...
+		},
+		Aggregation: &monitoringpb.Aggregation{
+			AlignmentPeriod:    duration.New(offset),
+			PerSeriesAligner:   monitoringpb.Aggregation_ALIGN_DELTA,
+			GroupByFields:      []string{"resource.labels.service_name"},
+			CrossSeriesReducer: monitoringpb.Aggregation_REDUCE_SUM,
+		},
+	})
+
+	for {
```
ahmetb, line 69: no need for a for loop if you get 1 data point or 1 page

internal/stackdriver/stackdriver.go
```diff
@@ -44,6 +44,46 @@ func NewAPIClient(ctx context.Context, project string) (*API, error) {
...
+		if err == iterator.Done {
+			return 0, nil
+		}
+		if err != nil {
+			return 0, errors.Wrap(err, "error when iterating through time series")
+		}
+
+		// The request count is aggregated for the entire revision, so only one
+		// point is returned.
+		return series.Points[0].Value.GetInt64Value(), nil
```
ahmetb, line 83: defensive code: instead of panicking if no data points come, check len==0 and return err

internal/stackdriver/stackdriver.go
```diff
@@ -44,6 +44,46 @@ func NewAPIClient(ctx context.Context, project string) (*API, error) {
...
 }
 
+// RequestCount count returns the number of requests for the given offset and
+// query.
+func (a *API) RequestCount(ctx context.Context, query Query, offset time.Duration) (int64, error) {
+	query = addFilterToQuery(query, "metric.type", requestCount)
+	endTime := time.Now()
+	startTime := endTime.Add(-1 * offset)
+
+	it := a.MetricClient.ListTimeSeries(ctx, &monitoringpb.ListTimeSeriesRequest{
```
ahmetb, line 54: use this as it doesn't return an iterator
https://godoc.org/google.golang.org/api/monitoring/v3#ProjectsTimeSeriesListCall.Do

internal/stackdriver/stackdriver.go
```diff
@@ -44,6 +44,46 @@ func NewAPIClient(ctx context.Context, project string) (*API, error) {
...
+			StartTime: timestamp.New(startTime),
+			EndTime:   timestamp.New(endTime),
+		},
+		Aggregation: &monitoringpb.Aggregation{
+			AlignmentPeriod:    duration.New(offset),
+			PerSeriesAligner:   monitoringpb.Aggregation_ALIGN_DELTA,
+			GroupByFields:      []string{"resource.labels.service_name"},
+			CrossSeriesReducer: monitoringpb.Aggregation_REDUCE_SUM,
+		},
+	})
```
ahmetb, line 67: need to handle ExecutionErrors field
https://godoc.org/google.golang.org/api/monitoring/v3#ListTimeSeriesResponse

it seems sometimes rpc succeeds, but api returns errors in a field.

ahmetb:
cc: @grayside if you have any experience with Stackdriver RPCs, might be good to take a look. otherwise no worries.

internal/stackdriver/stackdriver.go
gvso: My bad. I just noticed that this package is deprecated. That's why it looked a little bit weird to me yesterday. We are using [this one](https://pkg.go.dev/cloud.google.com/go/monitoring/apiv3?tab=doc) as recommended.

internal/stackdriver/stackdriver.go
gvso: With the new library, this is not the case (see my previous comment)

internal/stackdriver/stackdriver.go
ahmetb: If you think new pkg has all the capabilities use it. I suspect it's a handwritten pkg and we tend to have potential bugs or future deprecations as you've seen :)

internal/stackdriver/stackdriver.go
```diff
@@ -43,6 +44,45 @@ func NewAPIClient(ctx context.Context, project string) (*API, error) {
...
+		AggregationPerSeriesAligner("ALIGN_DELTA").
+		AggregationGroupByFields("resource.labels.service_name").
+		AggregationCrossSeriesReducer("REDUCE_SUM")
+
+	resp, err := req.Do()
+	if err != nil {
+		return 0, errors.Wrap(err, "error when retrieving time series")
+	}
+	if len(resp.ExecutionErrors) != 0 {
+		return 0, errors.Errorf("execution errors occurred: %v", resp.ExecutionErrors)
```
ahmetb, line 69: might be better to use %#v here OR actually loop over execution errors and log them as .warn

gvso: Done. The tests will fail until #22 is merged.

internal/stackdriver/stackdriver.go
```diff
@@ -43,6 +45,50 @@ func NewAPIClient(ctx context.Context, project string) (*API, error) {
...
+		AggregationCrossSeriesReducer("REDUCE_SUM")
+
+	logger.Debug("querying request count from Cloud Monitoring API")
+	resp, err := req.Do()
+	if err != nil {
+		return 0, errors.Wrap(err, "error when retrieving time series")
+	}
+	if len(resp.ExecutionErrors) != 0 {
+		for _, execError := range resp.ExecutionErrors {
+			loggerWith.WithField("message", execError.Message).Warn("execution error occurred")
```
ahmetb, line 73: I think this is meant to be called `logger`, so the github actions’ message is correct?

internal/stackdriver/stackdriver.go
```diff
@@ -43,6 +45,50 @@ func NewAPIClient(ctx context.Context, project string) (*API, error) {
...
+	req := a.MetricClient.Projects.TimeSeries.List("projects/" + a.Project).
+		Filter(query.Query()).
+		IntervalStartTime(startTime.Format(time.RFC3339Nano)).
+		IntervalEndTime(endTime.Format(time.RFC3339Nano)).
+		AggregationAlignmentPeriod(offsetString).
+		AggregationPerSeriesAligner("ALIGN_DELTA").
+		AggregationGroupByFields("resource.labels.service_name").
+		AggregationCrossSeriesReducer("REDUCE_SUM")
+
+	logger.Debug("querying request count from Cloud Monitoring API")
```
ahmetb, line 66: you might find it useful to add values here.
imagine we query latency for a service 3 times.
but you see same log 3 times, you won't show what the each query was for.
but if you add some specifics (e.g. reducer, or aligner), you can have a better time distinguishing them.

internal/stackdriver/stackdriver.go
gvso: Yeah, problems with my editor after the Crostini VM has been open for some time.

internal/stackdriver/stackdriver.go
gvso: Since reducer and aligner are very "abstract" concepts about how Cloud Monitoring works (and are always the same), what about using `intervalStartTime` and `intervalEndTime`, which are more useful and changed depending on when you call the function?

internal/metrics/metrics.go
```diff
@@ -24,14 +24,17 @@ const (
 )
 
 // Metrics represents a monitoring API such as Stackdriver.
-//
-// Latency returns the request latency after applying the specified series
-// aligner and cross series reducer. The result is in milliseconds.
-//
-// Error rate gets all the server responses. It calculates the error rate by
-// performing the operation (5xx responses / all responses).
 type Metrics interface {
```
grayside, line 27: From the standpoint of making the metrics in use flexible, I'm surprised to see `Metrics` as an interface.

I expected to see:

```
var metrics []Metric
type Metric interface
type RequestCount Metric
```

Will leave this as a future consideration, on second glance it looks like this was a previous decision :)

ahmetb: Yeah imo this should be named MetricsProvider, as it is meant to be an interface for Stackdriver and Prometheus like apis

gvso: Oh, I think the naming is what make it confusing. The interface is meant to facilitate the creation of different types of metrics providers (Stackdriver, Prometheus, etc). Might be worth renaming this interface. What do you think?

gvso: Seems like @ahmetb commented at the same time as me. I've created an issue about this #25 . I will rename this interface in a different PR

PR #14, opened 2020-07-21
run: Log when cached regions are used

gvso:
This lets the developer know that the list of regions were obtained from a cache. It also adds utility functions to set and get loggers to and from a context, which will also become useful when logging information about health and metrics.

### DIFF
--- a/internal/util/context.go
+++ b/internal/util/context.go
@@ -13,12 +13,16 @@ var loggerKey contextKeyLogger
 
 // ContextWithLogger returns a copy of the parent context that includes the
 // logger.
-func ContextWithLogger(ctx context.Context, logger *logrus.Logger) context.Context {
+func ContextWithLogger(ctx context.Context, logger *logrus.Entry) context.Context {
 	return context.WithValue(ctx, loggerKey, logger)
 }
 
 // LoggerFromContext returns the logger from the context.
-func LoggerFromContext(ctx context.Context) *logrus.Logger {
-	logger := ctx.Value(loggerKey).(*logrus.Logger)
+func LoggerFromContext(ctx context.Context) *logrus.Entry {
+	logger := ctx.Value(loggerKey).(*logrus.Entry)
+	if logger == nil {
+		logger = logrus.NewEntry(logrus.New())
+	}
+
 	return logger
 }
--- a/internal/util/context.go
+++ b/internal/util/context.go
@@ -6,21 +6,19 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
-type contextKey string
+type contextKeyLogger struct{}
 
-// Context keys.
-const (
-	contextKeyLogger = contextKey("logger")
-)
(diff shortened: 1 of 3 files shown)


internal/util/context.go
```diff
@@ -0,0 +1,26 @@
...
+
+	"github.com/sirupsen/logrus"
+)
+
+type contextKey string
+
+// Context keys.
+const (
+	contextKeyLogger = contextKey("logger")
+)
```
ahmetb, line 14: I actually had a better idea. You can do this:

`type contextKeyLogger struct {}`

and use `var key contextKeyLogger;` as the key.

internal/util/context.go
```diff
@@ -0,0 +1,26 @@
...
+
+// ContextWithLogger returns a copy of the parent context that includes the
+// logger.
+func ContextWithLogger(ctx context.Context, logger *logrus.Logger) context.Context {
+	return context.WithValue(ctx, contextKeyLogger, logger)
+}
+
+// LoggerFromContext returns the logger from the context.
+func LoggerFromContext(ctx context.Context) *logrus.Logger {
+	logger := ctx.Value(contextKeyLogger).(*logrus.Logger)
```
ahmetb, line 24: so we need to imagine nil scenario for this...

it's likely we'll make mistakes and send ctx without a logger.

so for nil, we should ideally return the default singleton logger from the logrus pkg, or something similar (that doesn't have fields etc)...

it would be easy to spot those stmts as they wouldn't have any fields. and we'd pass the proper ctx in that case.

gvso: Something I just noticed is that if we have something like
```
lg := logger.WithField("field", "value").Logger
ctx = util.ContextWithLogger(ctx, lg)
```

and

```
lg := util.LoggerFromContext(ctx)
```

the field is no respected. It looks like when doing `.Logger`, every field goes away. This is exactly the issue you had when you demoed logrus to me (if you remember).

The way to conserve the fields is by stroring `logrus.Entry` instead of `logrus.Logger`. Should we go with that one instead?

gvso: Mmm... Using `logrus.Entry` might also be a problem since there might be cases in which we don't want to specify fields. In those cases we cannot retrieve a `logrus.Entry` unless we do some sort of "hack" :/

ahmetb: yeah storing `logrus.Entry` internally in the context is not a problem, these libraries should be able to hide it.

internal/util/context.go
```diff
@@ -0,0 +1,28 @@
+package util
```
ahmetb, line 1: cc: @grayside can you lend us a review for this small file?
we've hacked a way to propagate the logger with the ctx.

grayside: 👍 LGTM

Maybe this should be package internal/logger?

gvso:
I have resigned the commits

PR #15, opened 2020-07-21
health: Create package to determine health

gvso:
This creates a new health package that has a main IsHealthy function to check if the metrics for the candidate revision meet the configuration criteria.

### DIFF
--- a/internal/metrics/metrics.go
+++ b/internal/metrics/metrics.go
@@ -48,7 +48,7 @@ func PercentileToAlignReduce(percentile float64) (AlignReduce, error) {
 		return Align99Reduce99, nil
 	case 95:
 		return Align95Reduce95, nil
-	case 50:
+	case 50.0:
 		return Align50Reduce50, nil
 	default:
 		return 0, errors.Errorf("unsupported percentile value %.2f", percentile)
--- a/internal/metrics/metrics.go
+++ b/internal/metrics/metrics.go
@@ -2,9 +2,8 @@ package metrics
 
 import (
 	"context"
+	"fmt"
 	"time"
-
-	"github.com/pkg/errors"
 )
 
 // Query is the filter used to retrieve metrics data.
@@ -51,6 +50,6 @@ func PercentileToAlignReduce(percentile float64) (AlignReduce, error) {
 	case 50:
 		return Align50Reduce50, nil
 	default:
-		return 0, errors.Errorf("unsupported percentile value %.2f", percentile)
+		return 0, fmt.Errorf("unsupported percentile value %.2f", percentile)
 	}
 }
--- a/internal/metrics/metrics.go
+++ b/internal/metrics/metrics.go
@@ -48,7 +48,7 @@ func PercentileToAlignReduce(percentile float64) (AlignReduce, error) {
 		return Align99Reduce99, nil
 	case 95:
 		return Align95Reduce95, nil
-	case 50.0:
+	case 50:
 		return Align50Reduce50, nil
 	default:
 		return 0, errors.Errorf("unsupported percentile value %.2f", percentile)
--- a/internal/metrics/metrics.go
+++ b/internal/metrics/metrics.go
@@ -2,8 +2,9 @@ package metrics
 
 import (
 	"context"
-	"fmt"
 	"time"
+
+	"github.com/pkg/errors"
 )
 
 // Query is the filter used to retrieve metrics data.
@@ -50,6 +51,6 @@ func PercentileToAlignReduce(percentile float64) (AlignReduce, error) {
 	case 50:
 		return Align50Reduce50, nil
 	default:
-		return 0, fmt.Errorf("unsupported percentile value %.2f", percentile)
+		return 0, errors.Errorf("unsupported percentile value %.2f", percentile)
 	}
 }
--- a/pkg/health/health.go
+++ b/pkg/health/health.go
@@ -35,16 +35,16 @@ type CheckResult struct {
 // Otherwise, all metrics criteria are checked to determine if the revision is
 // healthy.
 func Diagnose(ctx context.Context, provider metrics.Metrics, query metrics.Query,
-	offset time.Duration, minRequests int64, metricsCriteria []config.Metric) (*Diagnosis, error) {
+	offset time.Duration, minRequests int64, healthCriteria []config.Metric) (*Diagnosis, error) {
 
-	metricsValues, err := CollectMetrics(ctx, provider, query, offset, metricsCriteria)
+	metricsValues, err := CollectMetrics(ctx, provider, query, offset, healthCriteria)
 	if err != nil {
 		return nil, errors.Wrap(err, "could not collect metrics")
 	}
 
 	isHealthy := true
 	var results, failedResults []*CheckResult
-	for i, criteria := range metricsCriteria {
+	for i, criteria := range healthCriteria {
 		result := determineResult(criteria.Type, criteria.Min, criteria.Max, metricsValues[i])
 		results = append(results, result)
 
@@ -64,10 +64,10 @@ func Diagnose(ctx context.Context, provider metrics.Metrics, query metrics.Query
 
 // CollectMetrics returns an array of values collected for each of the specified
 // metrics criteria.
-func CollectMetrics(ctx context.Context, provider metrics.Metrics, query metrics.Query, offset time.Duration, metricsCriteria []config.Metric) ([]interface{}, error) {
+func CollectMetrics(ctx context.Context, provider metrics.Metrics, query metrics.Query, offset time.Duration, healthCriteria []config.Metric) ([]interface{}, error) {
 
 	var values []interface{}
-	for _, criteria := range metricsCriteria {
+	for _, criteria := range healthCriteria {
 		var value interface{}
 		var err error
 
--- a/pkg/health/health.go
+++ b/pkg/health/health.go
@@ -5,8 +5,10 @@ import (
 	"time"
 
 	"github.com/GoogleCloudPlatform/cloud-run-release-operator/internal/metrics"
+	"github.com/GoogleCloudPlatform/cloud-run-release-operator/internal/util"
 	"github.com/GoogleCloudPlatform/cloud-run-release-operator/pkg/config"
 	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
 )
 
 // DiagnosisResult is a possible result after a diagnosis.
@@ -43,6 +45,7 @@ type CheckResult struct {
 func Diagnose(ctx context.Context, provider metrics.Metrics, query metrics.Query,
 	offset time.Duration, minRequests int64, healthCriteria []config.Metric) (*Diagnosis, error) {
 
+	logger := util.LoggerFromContext(ctx)
 	metricsValues, err := CollectMetrics(ctx, provider, query, offset, healthCriteria)
 	if err != nil {
 		return nil, errors.Wrap(err, "could not collect metrics")
@@ -56,6 +59,17 @@ func Diagnose(ctx context.Context, provider metrics.Metrics, query metrics.Query
 
 		if !result.IsCriteriaMet {
 			overallResult = Unhealthy
+
+			logger := logger.WithFields(logrus.Fields{
+				"metricsType": criteria.Type,
+				"threshold":   criteria.Threshold,
+				"actualValue": result.ActualValue,
+			})
+			// If criteria is a latency, we want to log the percentile as well
+			if criteria.Type == config.LatencyMetricsCheck {
+				logger = logger.WithField("percentile", criteria.Percentile)
+			}
+			logger.Debug("criteria was not met")
 		}
 	}
 
@@ -68,6 +82,8 @@ func Diagnose(ctx context.Context, provider metrics.Metrics, query metrics.Query
 // CollectMetrics returns an array of values collected for each of the specified
 // metrics criteria.
 func CollectMetrics(ctx context.Context, provider metrics.Metrics, query metrics.Query, offset time.Duration, healthCriteria []config.Metric) ([]float64, error) {
+	logger := util.LoggerFromContext(ctx)
+	logger.Debug("start collecting metrics")
 	var values []float64
 	for _, criteria := range healthCriteria {
 		var value float64
@@ -117,21 +133,28 @@ func latency(ctx context.Context, provider metrics.Metrics, query metrics.Query,
 		return 0, errors.Wrap(err, "invalid percentile")
 	}
 
+	logger := util.LoggerFromContext(ctx).WithField("percentile", percentile)
+	logger.Debug("querying for latency metrics")
 	latency, err := provider.Latency(ctx, query, offset, alignerReducer)
 	if err != nil {
 		return 0, errors.Wrap(err, "failed to get latency metrics")
 	}
+	logger.WithField("value", latency).Debug("latency successfully retrieved")
 
 	return latency, nil
 }
 
 // errorRatePercent returns the percentage of errors during the given offset.
 func errorRatePercent(ctx context.Context, provider metrics.Metrics, query metrics.Query, offset time.Duration) (float64, error) {
+	logger := util.LoggerFromContext(ctx)
+	logger.Debug("querying for error rate metrics")
 	rate, err := provider.ErrorRate(ctx, query, offset)
 	if err != nil {
 		return 0, errors.Wrap(err, "failed to get error rate metrics")
 	}
 
 	// Multiply rate by 100 to have a percentage.
-	return rate * 100, nil
+	rate *= 100
+	logger.WithField("value", rate).Debug("error rate successfully retrieved")
+	return rate, nil
 }
--- a/pkg/health/health.go
+++ b/pkg/health/health.go
@@ -50,7 +50,12 @@ func (h *Health) IsHealthy(ctx context.Context, query metrics.Query, healthCheck
 
 // checkLatency checks that the threshold for latency is not exceeded.
 func (h *Health) checkLatency(ctx context.Context, query metrics.Query, offset time.Duration, percentile, max float64) (bool, error) {
-	latency, err := h.Metrics.Latency(ctx, query, offset, metrics.PercentileToAlignReduce(percentile))
+	alignerReducer, err := metrics.PercentileToAlignReduce(percentile)
+	if err != nil {
+		return false, errors.Wrap(err, "invalid percentile")
+	}
+
+	latency, err := h.Metrics.Latency(ctx, query, offset, alignerReducer)
 	if err != nil {
 		return false, errors.Wrap(err, "failed to get latency metrics")
 	}
--- a/pkg/health/health.go
+++ b/pkg/health/health.go
@@ -2,6 +2,7 @@ package health
 
 import (
 	"context"
+	"fmt"
 	"time"
 
 	"github.com/GoogleCloudPlatform/cloud-run-release-operator/internal/metrics"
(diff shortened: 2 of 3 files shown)


internal/metrics/metrics.go
```diff
@@ -32,3 +32,19 @@ type Metrics interface {
 	Latency(ctx context.Context, query Query, offset time.Duration, alignReduceType AlignReduce) (float64, error)
 	ErrorRate(ctx context.Context, query Query, offset time.Duration) (float64, error)
 }
+
+// PercentileToAlignReduce takes a percentile value maps it to a AlignReduce
+// value.
+//
+// TODO: once we start supporting any percentile value, this should not be
```
ahmetb, line 39: why are we not able to support floats right away?

gvso: Since we decided to use the REST API instead of MQL first, we don't need (and can't) support floats. The REST API, only allowed specific percentiles (99, 95, 50)

ahmetb: makes sense. let's have validation that if values like 97 is specified it should return error instead of defaulting to some unwanted value.

gvso: Done. I'm using `fmt.Errorf` since `pkg/errors` doesn't have a constructor with formatting.

ahmetb: don't use fmt.Errorf. Try errors.Errorf.

gvso: Will do so. That actually makes more sense. It's interesting that go-lint recommended `fmt.Errorf`

pkg/health/health.go
```diff
@@ -0,0 +1,75 @@
...
+)
+
+// Checker validates that revision is healthy.
+type Checker interface {
+	IsHealthy(ctx context.Context, query metrics.Query, healthCheckOffset time.Duration, metricsChecks []config.Metric) (bool, error)
+}
+
+// Health is the candidate's health status checker.
+type Health struct {
+	Metrics metrics.Metrics
```
ahmetb, line 19: unexport

pkg/health/health.go
```diff
@@ -0,0 +1,75 @@
...
+
+// New initializes a health checker.
+func New(metrics metrics.Metrics) *Health {
+	return &Health{
+		Metrics: metrics,
+	}
+}
+
+// IsHealthy checks if a revision is healthy based on the metrics configuration.
+func (h *Health) IsHealthy(ctx context.Context, query metrics.Query, healthCheckOffset time.Duration, metricsChecks []config.Metric) (bool, error) {
```
ahmetb, line 30: 1. Health doesn't have to be a type. You can pass `metrics.Metrics` here as well.
2. IsHealthy doesn't have to be defined on a type. see 1
3. You can easily test IsHealthy by passing mocks, no need for `interface Checker`

pkg/health/health.go
```diff
@@ -0,0 +1,75 @@
...
+	for _, check := range metricsChecks {
+		var err error
+		switch check.Type {
+		case config.LatencyMetricsCheck:
+			healthy, err = h.checkLatency(ctx, query, healthCheckOffset, check.Percentile, check.Max)
+			break
+		case config.ErrorRateMetricsCheck:
+			healthy, err = h.checkErrorRate(ctx, query, healthCheckOffset, check.Max)
+			break
+		}
```
ahmetb, line 41: default: return err unimplemented %s

pkg/health/health.go
ahmetb, line 30: Maybe we need to do this in another PR:

If you just return `bool` for fail/pass, we cannot capture details like "what metrics did we see", "why did we make this decision". 

As part of our design we said we'll provide these insights back to the users. So probably we need to export something like []CheckResults that explains result of each criteria checked and if it was "met" or "not".

internal/metrics/metrics.go
```diff
@@ -3,6 +3,8 @@ package metrics
 import (
```
gvso, line 3: Ignore this. Same as #17

internal/metrics/metrics_test.go
```diff
@@ -0,0 +1,73 @@
+package metrics_test
```
gvso, line 1: Ignore this. Same as #17

pkg/health/health.go
```diff
@@ -0,0 +1,158 @@
...
+}
+
+// Diagnose attempts to determine the health of a revision.
+//
+// If the minimum number of requests is not met, then health cannot be
+// determined and Diagnosis.EnoughRequests is set to false.
+//
+// Otherwise, all metrics criteria are checked to determine if the revision is
+// healthy.
+func Diagnose(ctx context.Context, provider metrics.Metrics, query metrics.Query,
```
gvso, line 37: This does not check request count yet since it depends on #13 . I will do it in a following PR

pkg/health/health.go
```diff
@@ -0,0 +1,158 @@
...
+	IsHealthy          bool
+}
+
+// CheckResult is information about a metrics criteria check.
+type CheckResult struct {
+	MetricsType   config.MetricsCheck
+	MinOrMaxValue interface{}
+	ActualValue   interface{}
+	IsCriteriaMet bool
+	Reason        string
```
ahmetb, line 27: it seems like MinOrMaxValue (rename: Threshold or ExpectedValue) and ActualValue are what we need in this type.

IsCriteriaMet is ok to.

but the type etc can be inferred from the []config.Metric we passed in.

ahmetb: also just use float64

pkg/health/health.go
```diff
@@ -0,0 +1,158 @@
...
+
+	metricsValues, err := CollectMetrics(ctx, provider, query, offset, healthCriteria)
+	if err != nil {
+		return nil, errors.Wrap(err, "could not collect metrics")
+	}
+
+	isHealthy := true
+	var results, failedResults []*CheckResult
+	for i, criteria := range healthCriteria {
+		result := determineResult(criteria.Type, criteria.Min, criteria.Max, metricsValues[i])
```
ahmetb, line 48: update config.Metric to have Threshold (not Min / Max) and here in this method we know if it should be used as ≤ or ≥

pkg/health/health.go
```diff
@@ -0,0 +1,158 @@
...
+		if !result.IsCriteriaMet {
+			isHealthy = false
+			failedResults = append(failedResults, result)
+		}
+	}
+
+	return &Diagnosis{
+		EnoughRequests:     true,
+		CheckResults:       results,
+		FailedCheckResults: failedResults,
```
ahmetb, line 60: no need for this

pkg/health/health.go
```diff
@@ -0,0 +1,158 @@
...
+		result := determineResult(criteria.Type, criteria.Min, criteria.Max, metricsValues[i])
+		results = append(results, result)
+
+		if !result.IsCriteriaMet {
+			isHealthy = false
+			failedResults = append(failedResults, result)
+		}
+	}
+
+	return &Diagnosis{
```
ahmetb, line 57: ideally this structure should also represent that no rollout operation should be done (halt/noop) based on minrequests.

similar to IsHealthy but different.

this is probably an enum DiagnosisResult { UNKNOWN=0 , INCONCLUSIVE, HEALTHY= , UNHEALTHY }

pkg/health/health.go
```diff
@@ -0,0 +1,158 @@
...
+		FailedCheckResults: failedResults,
+		IsHealthy:          isHealthy,
+	}, nil
+}
+
+// CollectMetrics returns an array of values collected for each of the specified
+// metrics criteria.
+func CollectMetrics(ctx context.Context, provider metrics.Metrics, query metrics.Query, offset time.Duration, healthCriteria []config.Metric) ([]interface{}, error) {
+
+	var values []interface{}
```
ahmetb, line 69: don't use empty interface

pkg/health/health.go
```diff
@@ -0,0 +1,158 @@
...
+	for _, criteria := range healthCriteria {
+		var value interface{}
+		var err error
+
+		switch criteria.Type {
+		case config.LatencyMetricsCheck:
+			value, err = latency(ctx, provider, query, offset, criteria.Percentile)
+			break
+		case config.ErrorRateMetricsCheck:
+			value, err = errorRate(ctx, provider, query, offset)
```
ahmetb, line 79: add logs

pkg/health/health.go
```diff
@@ -0,0 +1,158 @@
...
+	"github.com/GoogleCloudPlatform/cloud-run-release-operator/pkg/config"
+	"github.com/pkg/errors"
+)
+
+// Diagnosis is the information about the health of the revision.
+type Diagnosis struct {
+	EnoughRequests     bool
+	CheckResults       []*CheckResult
+	FailedCheckResults []*CheckResult
+	IsHealthy          bool
```
ahmetb, line 18: ideally this structure should also represent that no rollout operation should be done (halt/noop) based on minrequests.

similar to IsHealthy but different.

this is probably an enum DiagnosisResult { UNKNOWN=0 , INCONCLUSIVE, HEALTHY= , UNHEALTHY }

pkg/health/health.go
```diff
@@ -0,0 +1,158 @@
...
+	latency, err := provider.Latency(ctx, query, offset, alignerReducer)
+	if err != nil {
+		return 0, errors.Wrap(err, "failed to get latency metrics")
+	}
+
+	return latency, nil
+}
+
+// errorRate returns the percentage of errors during the given offset.
+func errorRate(ctx context.Context, provider metrics.Metrics, query metrics.Query, offset time.Duration) (float64, error) {
```
ahmetb, line 150: errorRatePercent

pkg/health/health.go
gvso: This will be done in a following PR after we integrate this package with the `rollout`

pkg/health/health.go
gvso: Done at #20

pkg/health/health.go
```diff
@@ -0,0 +1,160 @@
...
+	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
+)
+
+// DiagnosisResult is a possible result after a diagnosis.
+type DiagnosisResult int32
+
+// Possible diagnosis results.
+const (
+	Unknown      DiagnosisResult = 0
```
ahmetb, line 19: use iota syntax.

pkg/health/health.go
```diff
@@ -0,0 +1,160 @@
...
+
+	"github.com/GoogleCloudPlatform/cloud-run-release-operator/internal/metrics"
+	"github.com/GoogleCloudPlatform/cloud-run-release-operator/internal/util"
+	"github.com/GoogleCloudPlatform/cloud-run-release-operator/pkg/config"
+	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
+)
+
+// DiagnosisResult is a possible result after a diagnosis.
+type DiagnosisResult int32
```
ahmetb, line 15: why int32? can't this just be int

pkg/health/health.go
```diff
@@ -0,0 +1,160 @@
...
+	Unknown      DiagnosisResult = 0
+	Inconclusive DiagnosisResult = 1
+	Healthy      DiagnosisResult = 2
+	Unhealthy    DiagnosisResult = 3
+)
+
+// Diagnosis is the information about the health of the revision.
+type Diagnosis struct {
+	OverallResult DiagnosisResult
+	CheckResults  []*CheckResult
```
ahmetb, line 28: why is this pointer? use pointers very sparingly

pkg/health/health.go
```diff
@@ -0,0 +1,160 @@
...
+func Diagnose(ctx context.Context, provider metrics.Metrics, query metrics.Query,
+	offset time.Duration, minRequests int64, healthCriteria []config.Metric) (*Diagnosis, error) {
+
+	logger := util.LoggerFromContext(ctx)
+	metricsValues, err := CollectMetrics(ctx, provider, query, offset, healthCriteria)
+	if err != nil {
+		return nil, errors.Wrap(err, "could not collect metrics")
+	}
+
+	overallResult := Healthy
```
ahmetb, line 54: this is potentially a problem if no healthCriteria is specified and you return `Healthy`. that's why we define `Inconclusive` and `Unknown` for.

if you say `var overallResult DiagnosisResult` it will default to `Unknown`.

ahmetb: to be clear my concern is:  It's not defensive enough for a critical task such as rollout. user is much better off if we don't roll out or rollback, rather than blindly rolling forward due to a bug in our tool.

I need you to approach to the problem with this mindset.

pkg/health/health.go
```diff
@@ -0,0 +1,160 @@
...
+			overallResult = Unhealthy
+
+			logger := logger.WithFields(logrus.Fields{
+				"metricsType": criteria.Type,
+				"threshold":   criteria.Threshold,
+				"actualValue": result.ActualValue,
+			})
+			// If criteria is a latency, we want to log the percentile as well
+			if criteria.Type == config.LatencyMetricsCheck {
+				logger = logger.WithField("percentile", criteria.Percentile)
```
ahmetb, line 70: just log it always, it'll be a percentile=0 field that doesn't mean anything. it's cleaner that way (+log stms align in human-readable output)

pkg/health/health.go
```diff
@@ -0,0 +1,160 @@
...
+		return nil, errors.Wrap(err, "could not collect metrics")
+	}
+
+	overallResult := Healthy
+	var results []*CheckResult
+	for i, criteria := range healthCriteria {
+		result := determineResult(criteria.Type, criteria.Threshold, metricsValues[i])
+		results = append(results, result)
+
+		if !result.IsCriteriaMet {
```
ahmetb, line 60: I probably explained "avoid if-nesting" if you can more than a handful times now. :) 

What if you said:

```
for healthCriteria {

  if result.IsCriteriaMet {
    continue
  }

  // things you do if unhealthy
  // and you don't need to indent it
}
```

ahmetb: any time you write a lengthy `if { ... }` block I need you to keep this in mind: can you move the block outside the `if` somehow or not

pkg/health/health.go
```diff
@@ -0,0 +1,160 @@
...
+			logger := logger.WithFields(logrus.Fields{
+				"metricsType": criteria.Type,
+				"threshold":   criteria.Threshold,
+				"actualValue": result.ActualValue,
+			})
+			// If criteria is a latency, we want to log the percentile as well
+			if criteria.Type == config.LatencyMetricsCheck {
+				logger = logger.WithField("percentile", criteria.Percentile)
+			}
+			logger.Debug("criteria was not met")
```
ahmetb, line 72: why don't you initialize logger outside the if and also log `criteria met`

ahmetb:
cc: @grayside this is probably the most important PR of the project. If you have time, PTAL.

In this PR we create some core machinery to look at a set of metrics thresholds and look at Stackdriver Monitoring API and make a decision (roll forward/back, noop).

pkg/health/health_test.go
```diff
@@ -0,0 +1,128 @@
...
+func TestDiagnose(t *testing.T) {
+	metricsMock := &metricsMocker.Metrics{}
+	metricsMock.LatencyFn = func(ctx context.Context, query metrics.Query, offset time.Duration, alignReduceType metrics.AlignReduce) (float64, error) {
+		return 500, nil
+	}
+	metricsMock.ErrorRateFn = func(ctx context.Context, query metrics.Query, offset time.Duration) (float64, error) {
+		return 0.01, nil
+	}
+
+	tests := []struct {
```
grayside, line 24: It might be good to test metrics and thresholds set to 0. Even if that is working fine now, that seems likely to raise unexpected behavior in any future refactoring.

pkg/health/health.go
```diff
@@ -0,0 +1,160 @@
...
+// was met or not.
+func determineResult(metricsType config.MetricsCheck, threshold float64, actualValue float64) *CheckResult {
+	result := &CheckResult{ActualValue: actualValue, Threshold: threshold}
+
+	// As of now, the supported health criteria (latency and error rate) need to
+	// be less than the threshold. So, this is sufficient for now but might need
+	// to change to a switch statement when criteria with a minimum threshold is
+	// added.
+	if actualValue <= threshold {
+		result.IsCriteriaMet = true
```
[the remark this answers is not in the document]
ahmetb: @grayside so it's actually hard to look at `CheckResult` object and tell if it's a positive result or negative. 

e.g. for latency metrics, threshold=100 actual=200 → requirement is **failed**
however, for "min request" metrics threshold=100 actual=200 → requirement is **met**

the method that generates this CheckResult decides if those numbers mean something good or bad.

in case of "min request" metrics is **not met** → we simply do noop (wait more requests to happen)

however in case of "latency" metrics **not met** → means "rollback".

pkg/health/health.go
```diff
@@ -0,0 +1,160 @@
...
+}
+
+// Diagnose attempts to determine the health of a revision.
+//
+// If the minimum number of requests is not met, then health cannot be
+// determined and Diagnosis.EnoughRequests is set to false.
+//
+// Otherwise, all metrics criteria are checked to determine if the revision is
+// healthy.
+func Diagnose(ctx context.Context, provider metrics.Metrics, query metrics.Query,
```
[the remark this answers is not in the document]
ahmetb: I’ll try to explain the context here:

1. `provider` is a metrics backend that lets you query certain type of metrics. However, we specify "filters" such as "service_name=X,revision=Y,project=Z", and those (under the covers) might be implemented differently. 

    though I agree, actually we could make that as part of initialization process to the MetricsProvider (and initialize a different `provider` for each svc/revision/project/...)

2. `offset` is for being able to specify "metrics for last X minutes" (for example Stackdriver actually doesn't support last X minutes, rather, we need to compute startDate/endDate inside its metric provider impl).

3. `minRequests` is something we will (over time) transition to an actual `config.Metric` criteria (just like latency) so ideally this will go away from this method signature.

pkg/health/health.go
```diff
@@ -0,0 +1,160 @@
...
+		result.IsCriteriaMet = true
+	}
+	return result
+}
+
+// latency returns the latency for the given offset and percentile.
+func latency(ctx context.Context, provider metrics.Metrics, query metrics.Query, offset time.Duration, percentile float64) (float64, error) {
+	alignerReducer, err := metrics.PercentileToAlignReduce(percentile)
+	if err != nil {
+		return 0, errors.Wrap(err, "invalid percentile")
```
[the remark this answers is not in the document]
ahmetb: we get the actual percentile value from the user (e.g. p94) however the API might not support that. so if it's an unsupported value, we return a parse error here.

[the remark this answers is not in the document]
ahmetb: however I agree we should probably notice this "parse error" waaay early on, before even starting to diagnose the service.

pkg/health/health.go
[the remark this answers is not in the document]
grayside: Ah, that makes sense, and reading again I see the comment about the forthcoming switch statement. That seems like it will be a bit complex. I can't help but go to the idea that there should be an interface such as the following (reusing the same verbs being discussed above, but more locally this time):

```go
type Metric interface {
  Assess(context.Context, metrics.Metrics, time.Duration, []config.Metrics[]) (*CheckResult, error)
  Diagnose() bool
}
```

Then the individual Metrics can be implemented with a bit more isolation and testability.
The current structure seems like it will work okay for 2-3 metrics but start to become less manageable thereafter.

pkg/health/health.go
gvso: Thanks! My initial thought was that `Inconclusive` is the value returned when there are no enough requests, but that's still not implemented.

I was not sure if the best approach to empty criteria was roll forward (maybe the user didn't want health checks) or do nothing (`Unknown`/`Inconclusive`).

pkg/health/health.go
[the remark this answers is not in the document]
gvso: Thanks @grayside! I think I understand your concern, but I am a little bit confused about your expected interface for the caller of this interface.

The idea is that you get a `CheckResult` per metric criteria since the `CheckResult` contains the expected threshold, the actual value, and whether the criteria is met based on those previous two values.

1) Do you want to pass a single `config.Metric` and get a single `CheckResult`?
```
for criteria := range healthCriteria {
    results := append(results, Assess(ctx, metricsProvider, ....)) // ignore error
}
```

2) Do you want to pass an array of metrics criteria an get an array of `CheckResult`?

3) Something else?

Also, I don't see how this will prevent the `switch` when determining results since using comparing `actualValue <= threshold` or `threshold <= actualValue` depends on the metrics criteria type (unless we pass a comparison function, which sounds like an overkill here).

pkg/health/health.go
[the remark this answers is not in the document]
gvso: The config checks are done in `main`, though checks for health criteria has not been added yet (#27). This is done here as a defensive approach as we have done with many other things. I would keep the `float64` here since we would need it once we support any percentile value.

pkg/health/health.go
```diff
@@ -0,0 +1,160 @@
...
+			// If criteria is a latency, we want to log the percentile as well
+			if criteria.Type == config.LatencyMetricsCheck {
+				logger = logger.WithField("percentile", criteria.Percentile)
+			}
+			logger.Debug("criteria was not met")
+		}
+	}
+
+	return &Diagnosis{
+		OverallResult: overallResult,
```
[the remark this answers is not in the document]
gvso: I like this approach, what about doing this:

```
results := health.Assess(ctx, provider, query, offset, healthCriteria) // returns []health.CheckResult
diagnosis := health.Diagnose(healthCriteria, results)
```

The above approach works too, it just means an extra struct containing `[]CheckResult`

pkg/health/health.go
gvso: Nice! I didn't know about this.

pkg/health/health.go
ahmetb: > I was not sure if the best approach to empty criteria was roll forward

Right, it's not a good approach.

Inconclusive or Unknown is better. But best might be returning error if no criteria is specified (terminal error).

pkg/health/health.go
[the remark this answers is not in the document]
ahmetb: I'm a bit +1 on this being an overkill right now.
Realistically we'll only have `a≤b` or `b≤a` comparisons as far as metrics go.
So creating a `LatencyMetric` or `RequestCountMetric` types probably are overkill.

Though I agree, Adam’s approach long-term is definitely appealing. I'd say no need to refactor this right now.

pkg/health/health.go
```diff
@@ -0,0 +1,160 @@
...
+		CheckResults:  results,
+	}, nil
+}
+
+// CollectMetrics returns an array of values collected for each of the specified
+// metrics criteria.
+func CollectMetrics(ctx context.Context, provider metrics.Metrics, query metrics.Query, offset time.Duration, healthCriteria []config.Metric) ([]float64, error) {
+	logger := util.LoggerFromContext(ctx)
+	logger.Debug("start collecting metrics")
```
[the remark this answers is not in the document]
ahmetb: I don't know how OT would play into here. Are you talking about "querying OT" or "using OT to trace the operator tool itself"?

We would look metrics storages (Stackdriver API, Prometheus) and those don't know anything about OT when it comes to querying.

pkg/health/health.go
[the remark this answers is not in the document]
ahmetb: Yeah, SGTM. Please capture this in an issue and let's do it another PR.

ahmetb:
Let’s address the quick actionable feedback and open issues for outstanding long-term issues and send follow-up PRs. This CL is getting pretty beefy.

pkg/health/health.go
gvso: My bad. This was a simple block before the last change. I will take this into account when making changes too

pkg/health/health.go
gvso: Fixed

pkg/health/health.go
gvso: Fixed!

pkg/health/health.go
gvso: Fixed!

pkg/health/health.go
gvso: Let's address this at #30

pkg/health/health.go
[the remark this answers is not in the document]
gvso: Keeping track of this at #30

pkg/health/health_test.go
gvso: Keeping track at #31

gvso:
I have addressed the quick actionable feedback, and I am keeping track of the other issues at #30 and #31

PR #16, opened 2020-07-22
metrics: Add a mock for the Metrics interface

gvso:


gvso:
As we discussed, I've moved these changes to a different PR, so we can focus on the updates for the health package in PR #15

PR #17, opened 2020-07-22
metrics: Convert percentile to align/reduce value

gvso:
This adds a function to convert a supported percentile to an equivalent value for time series aligner/reducer

### DIFF
--- a/internal/metrics/metrics.go
+++ b/internal/metrics/metrics.go
@@ -3,6 +3,8 @@ package metrics
 import (
 	"context"
 	"time"
+
+	"github.com/pkg/errors"
 )
 
 // Query is the filter used to retrieve metrics data.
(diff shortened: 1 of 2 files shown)


gvso:
As we discussed, I've moved these changes to a different PR, so we can focus on the updates for the health package in PR #15

internal/metrics/metrics.go
```diff
@@ -32,3 +34,21 @@ type Metrics interface {
...
 	ErrorRate(ctx context.Context, query Query, offset time.Duration) (float64, error)
 }
+
+// PercentileToAlignReduce takes a percentile value maps it to a AlignReduce
+// value.
+//
+// TODO: once we start supporting any percentile value, this should not be
+// needed.
+func PercentileToAlignReduce(percentile float64) (AlignReduce, error) {
+	switch percentile {
```
ahmetb, line 44: @broady kinda random but because of this https://play.golang.org/p/3of70nwGNo-
do you recommend we use `if .. else if` with `>=`

internal/metrics/metrics_test.go
```diff
@@ -0,0 +1,73 @@
...
+		},
+	}
+
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			alignReducer, err := metrics.PercentileToAlignReduce(test.percentile)
+			if test.shouldErr {
+				if err == nil {
+					fmt.Println(alignReducer)
+				}
```
ahmetb, line 66: seems like leftover

PR #19, opened 2020-07-23
stackdriver: Use package with RPC for MQL query

gvso:
This uses the Cloud Monitoring package that has support for MQL queries, which will be useful when we start supporting any percentile for latency. The package used here is deprecated, so we should migrate to the new package once support for MQL queries is added to the new Cloud Monitoring package.

### DIFF
--- a/internal/stackdriver/stackdriver.go
+++ b/internal/stackdriver/stackdriver.go
@@ -5,19 +5,18 @@ import (
 	"fmt"
 	"time"
 
-	monitoring "cloud.google.com/go/monitoring/apiv3"
 	"github.com/GoogleCloudPlatform/cloud-run-release-operator/internal/metrics"
 	"github.com/pkg/errors"
-	"google.golang.org/api/iterator"
-	monitoringpb "google.golang.org/genproto/googleapis/monitoring/v3"
-	duration "google.golang.org/protobuf/types/known/durationpb"
-	timestamp "google.golang.org/protobuf/types/known/timestamppb"
+
+	// TODO: Migrate to cloud.google.com/go/monitoring/apiv3/v2 once RPC for MQL
+	// query is added (https://cloud.google.com/monitoring/api/ref_v3/rest/v3/projects.timeSeries/query).
+	monitoring "google.golang.org/api/monitoring/v3"
 )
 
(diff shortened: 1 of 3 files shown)


internal/stackdriver/stackdriver.go
```diff
@@ -48,68 +47,67 @@ func NewAPIClient(ctx context.Context, project string) (*API, error) {
...
-	for {
-		series, err := it.Next()
-		if err == iterator.Done {
-			break
-		}
-		if err != nil {
-			return 0, errors.Wrap(err, "error when iterating through time series")
-		}
-
+	for _, series := range timeSeries {
```
ahmetb, line 105: don't do break, instead look at len(list), if 0, return err, then just process arr[0]

internal/stackdriver/stackdriver.go
```diff
@@ -48,68 +47,67 @@ func NewAPIClient(ctx context.Context, project string) (*API, error) {
...
 		// Because the interval and the series aligner are the same, only one
 		// point is returned per time series.
 		if series.Metric.Labels["response_code_class"] == codeClass {
-			latency = series.Points[0].Value.GetDoubleValue()
+			latency = *(series.Points[0].Value.DoubleValue)
```
ahmetb, line 109: we can be more defensive by adding [0] index check

internal/stackdriver/stackdriver.go
```diff
@@ -48,68 +47,67 @@ func NewAPIClient(ctx context.Context, project string) (*API, error) {
...
-		Aggregation: &monitoringpb.Aggregation{
-			AlignmentPeriod:    duration.New(offset),
-			PerSeriesAligner:   aligner,
-			GroupByFields:      []string{"metric.labels.response_code_class"},
-			CrossSeriesReducer: reducer,
-		},
-	})
-
-	return latencyForCodeClass(it, "2xx")
+	return latencyForCodeClass(resp.TimeSeries, "2xx")
```
ahmetb, line 71: we should take all latency classes into action not just 2xx

ahmetb: let's address that in another PR

PR #22, opened 2020-07-24
util: Defend against panic when logger is nil

gvso:
In a previous commit, a nil logger was supposed to be handled. However, the order of the check was reverted. This fixes this issue and handle cases in which the context does not have a logger value.

### DIFF
--- a/internal/util/context.go
+++ b/internal/util/context.go
@@ -19,10 +19,10 @@ func ContextWithLogger(ctx context.Context, logger *logrus.Entry) context.Contex
 
 // LoggerFromContext returns the logger from the context.
 func LoggerFromContext(ctx context.Context) *logrus.Entry {
-	logger := ctx.Value(loggerKey).(*logrus.Entry)
+	logger := ctx.Value(loggerKey)
 	if logger == nil {
 		logger = logrus.NewEntry(logrus.New())
 	}
 
-	return logger
+	return logger.(*logrus.Entry)
 }
--- a/internal/util/context.go
+++ b/internal/util/context.go
@@ -20,14 +20,9 @@ func ContextWithLogger(ctx context.Context, logger *logrus.Entry) context.Contex
 // LoggerFromContext returns the logger from the context.
 func LoggerFromContext(ctx context.Context) *logrus.Entry {
 	value := ctx.Value(loggerKey)
-	if value == nil {
-		logrus.Warnf("nil logger in context, creating a new one")
-		return logrus.NewEntry(logrus.New())
-	}
-
 	logger, ok := value.(*logrus.Entry)
 	if !ok {
-		logrus.Warnf("received wrong type of logger (%T)", logger)
+		logrus.Warnf("received wrong type of logger (%T)", value)
 		logger = logrus.NewEntry(logrus.New())
 	}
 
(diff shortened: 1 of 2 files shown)


internal/util/context.go
```diff
@@ -19,10 +19,10 @@ func ContextWithLogger(ctx context.Context, logger *logrus.Entry) context.Contex
...
 // LoggerFromContext returns the logger from the context.
 func LoggerFromContext(ctx context.Context) *logrus.Entry {
-	logger := ctx.Value(loggerKey).(*logrus.Entry)
+	logger := ctx.Value(loggerKey)
 	if logger == nil {
 		logger = logrus.NewEntry(logrus.New())
 	}
 
-	return logger
+	return logger.(*logrus.Entry)
```
ahmetb, line 27: I think this is still wrong and could still panic (what if we don't get *logrus.Entry type, but something else?).

Do this:

```
v, ok := logger.(*logrus.Entry)
if !ok {
	logrus.Warnf("received wrong type of logger (%T)", logger)
    // initialize default logger as above
}
```

gvso: I added check for nil and for wrong type. I also created tests for this.

I kept check for nil because when nil, the error message was ""received wrong type of logger (*logrus.Entry)", which is what we actually want and might be confusing.

ahmetb: > the error message was "received wrong type of logger (*logrus.Entry)", which is what we actually want and might be confusing.

these error messages help us as developers to fix it.
but it is meant to prevent the process from crashing so users can continue to use it if we accidentally create a bug by passing loggerless context.

internal/util/context.go
```diff
@@ -19,8 +19,15 @@ func ContextWithLogger(ctx context.Context, logger *logrus.Entry) context.Contex
 
 // LoggerFromContext returns the logger from the context.
 func LoggerFromContext(ctx context.Context) *logrus.Entry {
```
ahmetb, line 21: in the long run we can rename this to `LoggerFrom` (as it's clear it's from  ctx).

open an issue to track pls.

internal/util/context.go
```diff
@@ -19,8 +19,15 @@ func ContextWithLogger(ctx context.Context, logger *logrus.Entry) context.Contex
...
-	if logger == nil {
+	value := ctx.Value(loggerKey)
+	if value == nil {
+		logrus.Warnf("nil logger in context, creating a new one")
+		return logrus.NewEntry(logrus.New())
+	}
+
+	logger, ok := value.(*logrus.Entry)
+	if !ok {
+		logrus.Warnf("received wrong type of logger (%T)", logger)
```
ahmetb, line 30: %T on logger won't work
%T should be passed `value`

internal/util/context.go
```diff
@@ -19,8 +19,15 @@ func ContextWithLogger(ctx context.Context, logger *logrus.Entry) context.Contex
...
 		logger = logrus.NewEntry(logrus.New())
```
ahmetb, line 31: you should learn by now, this is code repetition. you have 2 branches that do the same thing.

for example, if you tried it yourself instead of directly accepting my suggestion you'd find that nil actually works with `v, ok := ...(type)` assertion just fine.

https://play.golang.org/p/Nt2AyjMiqwG

PR #29, opened 2020-07-27
metrics: Rename interface to Provider

gvso:
The current interface name, Metrics, is confusing since it doesn't reflect what the interface is about. It looks like the interface refers to a specific metrics criteria while in reality represents a metrics provider that returns metrics values.


ahmetb:
Sorry needs rebase.

gvso:
No problem! I was expecting that

PR #32, opened 2020-07-27
metrics: Remove the Query interface

gvso:
The Query interface was always passed through the chain without changes to it, which makes it very hard to deduct its use. This removes the interface altogether and keep the filters as parameters of the provider constructor.

A new method SetCandidateRevision was added to the Provider interface since the candidate revision can only be determined during the rollout process. Furthermore, it looks like knowing the queried resource (i.e. the candidate revision) is necessary for any metrics provider.

### DIFF
--- a/internal/metrics/metrics.go
+++ b/internal/metrics/metrics.go
@@ -23,6 +23,8 @@ const (
 type Provider interface {
 	// Sets the candidate revision name for which the provider should get
 	// metrics.
+	// TODO: Consider removing this method and making revisionName part of other
+	// method signatures.
 	SetCandidateRevision(revisionName string)
 
 	// Returns the number of requests for the given offset and query.
--- a/internal/metrics/metrics.go
+++ b/internal/metrics/metrics.go
@@ -7,34 +7,33 @@ import (
 	"github.com/pkg/errors"
 )
 
-// Query is the filter used to retrieve metrics data.
-type Query interface {
-	Query() string
-}
-
 // AlignReduce is the type to enumerate allowed combinations of per series
 // aligner and cross series reducer.
 type AlignReduce int32
 
 // Series aligner and cross series reducer types (for latency).
 const (
-	Align99Reduce99 AlignReduce = 1
-	Align95Reduce95             = 2
-	Align50Reduce50             = 3
+	Align99Reduce99 AlignReduce = iota + 1
+	Align95Reduce95
+	Align50Reduce50
 )
 
 // Provider represents a metrics Provider such as Stackdriver.
 type Provider interface {
+	// Sets the candidate revision name for which the provider should get
+	// metrics.
+	SetCandidateRevision(revisionName string)
+
 	// Returns the number of requests for the given offset and query.
-	RequestCount(ctx context.Context, query Query, offset time.Duration) (int64, error)
+	RequestCount(ctx context.Context, offset time.Duration) (int64, error)
 
 	// Returns the request latency after applying the specified series aligner
 	// and cross series reducer. The result is in milliseconds.
-	Latency(ctx context.Context, query Query, offset time.Duration, alignReduceType AlignReduce) (float64, error)
+	Latency(ctx context.Context, offset time.Duration, alignReduceType AlignReduce) (float64, error)
 
 	// Gets all the server responses and calculates the error rate by performing
 	// the operation (5xx responses / all responses).
-	ErrorRate(ctx context.Context, query Query, offset time.Duration) (float64, error)
+	ErrorRate(ctx context.Context, offset time.Duration) (float64, error)
 }
 
 // PercentileToAlignReduce takes a percentile value maps it to a AlignReduce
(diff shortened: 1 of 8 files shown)


internal/stackdriver/stackdriver.go
```diff
@@ -228,31 +237,22 @@ func alignerAndReducer(alignReduceType metrics.AlignReduce) (aligner string, red
...
 
 // addFilter adds a filter to the query.
-func (q Query) addFilter(key, value string) Query {
+//
+// TODO: Support field-based filters, so the query string is generated based on
+// the fields instead of appending a filter everytime this method is called.
+func (q query) addFilter(key, value string) query {
 	if q.filter != "" {
 		q.filter += " AND "
 	}
```
ahmetb, line 254: doesn't the query end with `" AND"` in this case?

gvso: No, this case is only when you're adding the first filter. You can see that right after this conditional, the filter is appended

```go
q.filter += fmt.Sprintf("%s=%q", key, value)
```

Let me add a comment in the if statement to make it more clear the case it covers.

ahmetb: my bad don't worry

internal/stackdriver/stackdriver.go
```diff
@@ -12,19 +12,21 @@ import (
...
 	// query is added (https://cloud.google.com/monitoring/api/ref_v3/rest/v3/projects.timeSeries/query).
-
 	monitoring "google.golang.org/api/monitoring/v3"
 )
 
+// query is the filter used to retrieve metrics data.
+type query struct {
+	filter string
+}
+
```
ahmetb, line 22: can you `type query string`

gvso: Done! Can you confirm if the best way to pass a string to the Cloud Monitoring package is type conversion `string(query)`?

internal/stackdriver/stackdriver_internal_test.go
```diff
@@ -0,0 +1,39 @@
+package stackdriver
```
ahmetb, line 1: File name should be x_test.go. Internal is extra

internal/stackdriver/stackdriver_internal_test.go
```diff
@@ -0,0 +1,39 @@
...
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestAddFilter(t *testing.T) {
```
ahmetb, line 9: TestQuery_addFilter is a bit better.

internal/stackdriver/stackdriver_internal_test.go
```diff
@@ -0,0 +1,39 @@
...
+			name:     "single filter",
+			keys:     []string{"key1"},
+			values:   []string{"value1"},
+			expected: "key1=\"value1\"",
+		},
+		{
+			name:     "multiple filters",
+			keys:     []string{"key1", "key2", "key3"},
+			values:   []string{"value1", "value2", "value3"},
+			expected: "key1=\"value1\" AND key2=\"value2\" AND key3=\"value3\"",
```
ahmetb, line 26: Use backtick syntax for readability

internal/metrics/metrics.go
```diff
@@ -7,34 +7,33 @@ import (
...
 // AlignReduce is the type to enumerate allowed combinations of per series
 // aligner and cross series reducer.
 type AlignReduce int32
 
 // Series aligner and cross series reducer types (for latency).
 const (
-	Align99Reduce99 AlignReduce = 1
-	Align95Reduce95             = 2
-	Align50Reduce50             = 3
+	Align99Reduce99 AlignReduce = iota + 1
```
ahmetb, line 16: Again we can have "Unknown iota" as first.

internal/metrics/metrics.go
```diff
@@ -7,34 +7,33 @@ import (
...
+	Align95Reduce95
+	Align50Reduce50
 )
 
 // Provider represents a metrics Provider such as Stackdriver.
 type Provider interface {
+	// Sets the candidate revision name for which the provider should get
+	// metrics.
+	SetCandidateRevision(revisionName string)
```
ahmetb, line 25: Why is this not part of constructor of each provider? I don't see a need to change interface. The same can apply to other filters, but you have revisionName singled out in a new method?

gvso: We are injecting the metrics provider into the `rollout.Rollout` struct when we initialize it with `rollout.New`. At this point, we still don't know the candidate revision since it is determine during the [rollout process](https://github.com/GoogleCloudPlatform/cloud-run-release-operator/blob/main/pkg/rollout/rollout.go#L93). 

If we were to remove this function from the interface, we'd have to do all that logic about getting the stable and candidate revisions before initializing `rollout.Rollout`.

ahmetb: Let's add a todo like:

> TODO(gvso) Consider removing this method and making revisionName part of other method signatures.

gvso: Done!

PR #33, opened 2020-07-28
main: Separate rollout handling logic

gvso:
Most of runCLI is code that should be useful when handling rollouts in other contexts (e.g. http request). This separates that logic into a new function, which would also be useful when implementing concurrent rollouts.

### DIFF
--- a/cmd/operator/main.go
+++ b/cmd/operator/main.go
@@ -17,7 +17,6 @@ package main
 import (
 	"context"
 	"flag"
-	"log"
 	"os"
 	"strconv"
 	"strings"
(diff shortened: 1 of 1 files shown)


cmd/operator/main.go
```diff
@@ -154,25 +148,43 @@ func runCLI(ctx context.Context, logger *logrus.Logger, metricsProvider metrics.
...
+	}
+	roll := rollout.New(client, metricsProvider, service, strategy).WithLogger(lg.Logger)
+
+	changed, err := roll.Rollout()
+	if err != nil {
+		lg.Errorf("rollout failed: %v", err)
+	}
+
+	if changed {
+		lg.Debug("service was successfully updated")
```
ahmetb, line 182: this is less likely to happen.
so I'd log this as info (user visible by def)

cmd/operator/main.go
```diff
@@ -154,25 +148,43 @@ func runCLI(ctx context.Context, logger *logrus.Logger, metricsProvider metrics.
...
+		"region":  service.Region,
+	})
+
+	client, err := runapi.NewAPIClient(ctx, service.Region)
+	if err != nil {
+		lg.Error("failed to initialize Cloud Run API client")
+	}
+	metricsProvider, err := stackdriver.NewProvider(ctx, flProject)
+	if err != nil {
+		logger.Fatalf("failed to initialize metrics provider: %v", err)
```
ahmetb, line 172: don't do fatals

cmd/operator/main.go
```diff
@@ -154,25 +148,43 @@ func runCLI(ctx context.Context, logger *logrus.Logger, metricsProvider metrics.
...
+func handleRollout(ctx context.Context, logger *logrus.Logger, service *rollout.ServiceRecord, strategy *config.Strategy) {
+	lg := logger.WithFields(logrus.Fields{
+		"project": service.Project,
+		"service": service.Metadata.Name,
+		"region":  service.Region,
+	})
+
+	client, err := runapi.NewAPIClient(ctx, service.Region)
+	if err != nil {
+		lg.Error("failed to initialize Cloud Run API client")
```
ahmetb, line 168: return err

cmd/operator/main.go
```diff
@@ -154,25 +148,43 @@ func runCLI(ctx context.Context, logger *logrus.Logger, metricsProvider metrics.
...
-		}
+		handleRollout(ctx, logger, services[0], cfg.Strategy)
 
 		duration := time.Duration(cfg.Strategy.Interval)
 		time.Sleep(duration * time.Second)
 	}
 }
 
+// handleRollout manages the rollout process for a single service.
+func handleRollout(ctx context.Context, logger *logrus.Logger, service *rollout.ServiceRecord, strategy *config.Strategy) {
```
ahmetb, line 159: return err  from here.

cmd/operator/main.go
```diff
@@ -154,25 +148,43 @@ func runCLI(ctx context.Context, logger *logrus.Logger, metricsProvider metrics.
...
+	}
+	metricsProvider, err := stackdriver.NewProvider(ctx, flProject)
+	if err != nil {
+		logger.Fatalf("failed to initialize metrics provider: %v", err)
+	}
+	roll := rollout.New(client, metricsProvider, service, strategy).WithLogger(lg.Logger)
+
+	changed, err := roll.Rollout()
+	if err != nil {
+		lg.Errorf("rollout failed: %v", err)
```
ahmetb, line 178: return err

PR #34, opened 2020-07-28
health: Separate metrics collection from diagnosis

### LINKED ISSUE #30: health: Separate metrics collection from diagnosis
Let's separate the collection of metrics values from the diagnosis of the revision's health as discussed [here](https://github.com/GoogleCloudPlatform/cloud-run-release-operator/pull/15#discussion_r460333117)

gvso:
This separates the collection of metrics values from the actual diagnosis of the health of the revision. It closes #30

### DIFF
--- a/pkg/health/health.go
+++ b/pkg/health/health.go
@@ -11,79 +11,65 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
-// DiagnosisResult is a possible result after a diagnosis.
-type DiagnosisResult int
+// Diagnosis is a possible result after a diagnosis.
+type Diagnosis int
 
 // Possible diagnosis results.
 const (
-	Unknown DiagnosisResult = iota
+	Unknown Diagnosis = iota
 	Inconclusive
 	Healthy
 	Unhealthy
 )
 
-// Diagnosis is the information about the health of the revision.
-type Diagnosis struct {
-	OverallResult DiagnosisResult
-	CheckResults  []CheckResult
-}
-
-// CheckResult is information about a metrics criteria check.
-type CheckResult struct {
-	Threshold     float64
-	ActualValue   float64
-	IsCriteriaMet bool
-}
-
 // Diagnose attempts to determine the health of a revision.
 //
-// If the minimum number of requests is not met, then health cannot be
-// determined and Diagnosis.EnoughRequests is set to false.
+// If no health criteria is specified or the size of the health criteria and the
+// actual values are not the same, the diagnosis is Unknown and an error is
+// returned.
 //
-// Otherwise, all metrics criteria are checked to determine if the revision is
-// healthy.
-func Diagnose(ctx context.Context, provider metrics.Provider, offset time.Duration,
-	minRequests int64, healthCriteria []config.Metric) (*Diagnosis, error) {
-
+// If the minimum number of requests is not met, then health cannot be
+// determined and diagnosis is Inconclusive.
+//
+// Otherwise, all metrics criteria are checked to determine whether the revision
+// is healthy or not.
+func Diagnose(ctx context.Context, healthCriteria []config.Metric, actualValues []float64) (Diagnosis, error) {
 	logger := util.LoggerFromContext(ctx)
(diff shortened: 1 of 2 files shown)


cmd/operator/main.go
```diff
@@ -23,7 +23,6 @@ import (
 	"strings"
```
gvso, line 23: Ignore. Same as #32

internal/metrics/metrics.go
```diff
@@ -7,34 +7,33 @@ import (
 	"github.com/pkg/errors"
```
gvso, line 7: Ignore. Same as #32

internal/metrics/mock/metrics.go
```diff
@@ -9,35 +9,46 @@ import (
```
gvso, line 9: Ignore. Same as #32

internal/stackdriver/stackdriver.go
```diff
@@ -12,19 +12,21 @@ import (
```
gvso, line 12: Ignore. Same as #32

pkg/health/health.go
```diff
@@ -11,89 +11,75 @@ import (
...
+		return Unknown, errors.New("the size of health criteria is not the same to the size of the actual metrics values")
+	}
+	if len(healthCriteria) == 0 {
+		return Unknown, errors.New("health criteria must be specified")
 	}
 
-	overallResult := Healthy
-	var results []CheckResult
-	for i, criteria := range healthCriteria {
+	diagnosis := Healthy
```
gvso, line 45: Still not handling `Inconclusive`. It will be added later, see #35

pkg/health/health.go
```diff
@@ -11,89 +11,75 @@ import (
...
-		overallResult = Unhealthy
-		logger.Debug("unmet criteria")
+		logger.Debug("met criterion")
 	}
 
-	return &Diagnosis{
-		OverallResult: overallResult,
-		CheckResults:  results,
-	}, nil
+	return diagnosis, nil
```
ahmetb, line 63: let's bring this type back

```
// Diagnosis is the information about the health of the revision.	
type Diagnosis struct {	
	OverallResult DiagnosisResult	
	CheckResults  []CheckResult	
}	
```

pkg/health/health.go
ahmetb: Consider this approach.
https://play.golang.org/p/ikdSjj1F0ae

Here I define a bitmap and mark if I've seen a "inconclusive" or "unhealthy". At the end I make a decision.

gvso: This fails when we remove the `Inconclusive` result https://play.golang.org/p/IcOr8bjz_4L
The problem is that `Unhealthy` gets 6. If we want the bitmap to work, we need the values to be [1, 2, 4, 8] or [0, 2, 4, 8] instead of [0, 2, 4, 6]
I think it was just a typo in your example since this works https://play.golang.org/p/yym1xTQD9y-

I'm still a little worried about readability though.

gvso: Ok. Thinking about it. If we look at the conditionals at https://play.golang.org/p/yym1xTQD9y-, we can see that priorities are set. That is, we care about results in the following order
1) Are there any inconclusive results?
2) Are there any unmet criteria?
3) Are all criteria met?
4) Unknown

Notice that unknown can only happen if no 1-3 do not, but we are already checking that you have at least a metrics criteria, so 1-3 must happen if you get to that point. 

We can also just start with Unknown and set the health to `Healthy` only the first time a criteria is met:
```
// Criteria is met at this point
if diagnosis == Unknown {
   diagnosis = Healthy
}
```

We will have a special case for request count anyways, so we can break at that point if no enough requests were made and return `Inconclusive`

With this approach, we're basically maintaining the same priority as with the bitmap, and in my view, we make it simpler to read.

ahmetb: OK sounds good. We don't have to initialize as Unknown either if you think the code is error-proof for future iterations (potentially by others).

gvso: Still want to initialize as `Unknown` just in case the only health criteria is request count.

PR #37, opened 2020-07-28
rollout: Add methods to include client and context

gvso:
This adds two methods that allows the update of the client and context used for rollouts. The context would become useful when we have to pass a context to the health package.

### DIFF
--- a/pkg/rollout/rollout.go
+++ b/pkg/rollout/rollout.go
@@ -54,6 +54,7 @@ func New(metricsProvider metrics.Provider, svcRecord *ServiceRecord, strategy *c
 		project:         svcRecord.Project,
 		region:          svcRecord.Region,
 		strategy:        strategy,
+		ctx:             context.TODO(),
 		log:             logrus.NewEntry(logrus.New()),
 	}
 }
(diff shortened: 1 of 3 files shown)


gvso:
@ahmetb  We discussed about adding at `WithClient` function. I went ahead and added a `WithContext` function too.

pkg/rollout/rollout.go
```diff
@@ -41,22 +43,33 @@ const (
...
 }
 
+// WithClient updates the client in the rollout instance.
+func (r *Rollout) WithClient(client runapi.Client) *Rollout {
+	r.runClient = client
+	return r
+}
+
+// WithContext updates the context in the rollout instance.
+func (r *Rollout) WithContext(ctx context.Context) *Rollout {
```
ahmetb, line 68: I don't get WithContext.
You can technically always pass context.TODO() or context.Background() for tests?
What's the purpose of this?

gvso: That's true, but we also wanted to reduce the number of parameters in the constructor. 

However, it doesn't hurt to add it to `New`

ahmetb: ctx is typically part of method signatures.

PR #40, opened 2020-07-30
config: Validate metrics criteria configuration

gvso:
This validates that the metrics criteria to err sooner for an invalid value. It also adds an extra return value in `config.IsValid` to specify the problem with the configuration object.

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -1,12 +1,16 @@
 package config
 
+import (
+	"github.com/pkg/errors"
+)
+
 // MetricsCheck is the metrics check type.
 type MetricsCheck string
 
 // Supported metrics checks.
 const (
 	LatencyMetricsCheck   MetricsCheck = "request-latency"
-	ErrorRateMetricsCheck              = "error-rate-percent"
+	ErrorRateMetricsCheck MetricsCheck = "error-rate-percent"
 )
 
 // Target is the configuration to filter services.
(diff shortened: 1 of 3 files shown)


pkg/config/config.go
```diff
@@ -72,38 +76,66 @@ func NewTarget(project string, regions []string, labelSelector string) *Target {
...
 	var previous int64
 	for _, step := range config.Strategy.Steps {
 		if step <= previous || step > 100 {
-			return false
+			return false, errors.New("steps must be in ascending order and not greater than 100")
 		}
 		previous = step
 	}
 
+	for _, criteria := range config.Strategy.Metrics {
```
ahmetb, line 97: add `for i, ` and add %d with `i` to the error

pkg/config/config.go
```diff
@@ -72,38 +76,66 @@ func NewTarget(project string, regions []string, labelSelector string) *Target {
...
 
-func targetsAreValid(targets []*Target) bool {
+func metricsIsValid(metricsCriteria Metric) (bool, error) {
+	threshold := metricsCriteria.Threshold
+	switch metricsCriteria.Type {
+	case ErrorRateMetricsCheck:
+		if threshold > 100 || threshold < 0 {
+			return false, errors.Errorf("threshold must be greater than 0 and less than 100 for %q", metricsCriteria.Type)
+		}
+		break
```
ahmetb, line 112: `break` not necessary

PR #41, opened 2020-07-30
main: Make concurrent rollouts

gvso:
This concurrently rollouts all the services retrieved from the API.

### DIFF
--- a/cmd/operator/main.go
+++ b/cmd/operator/main.go
@@ -17,9 +17,11 @@ package main
 import (
 	"context"
 	"flag"
+	"fmt"
 	"os"
 	"strconv"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/GoogleCloudPlatform/cloud-run-release-operator/internal/metrics"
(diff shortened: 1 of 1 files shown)


cmd/operator/main.go
```diff
@@ -155,11 +157,30 @@ func runDaemon(ctx context.Context, logger *logrus.Logger, cfg *config.Config) e
...
+		wg.Wait()
+
+		var errsStr string
+		for i, err := range errs {
+			errsStr += fmt.Sprintf("\n[error#%d] %v", i, err)
+		}
+		if len(errs) != 0 {
+			logger.Warnf("there were %d errors: %s", len(errs), errsStr)
 		}
+
```
gvso, line 183: I'm not returning an error here since a failure to update service 'A' shouldn't finish the program. Yet maybe an error when creating a metrics provider should be fatal (since it might be common across all the services). Should we add a specific check if one of the errors is about creating metrics provider?

ahmetb: So this depends on the execution mode. instead of the `logger.Warnf` above

* if this tool runs "once" with request-trigger on Cloud Run: we need to respond to the request with HTTP 500 + this compiled error message

* if this tool runs continuously (e.g. invoked from shell): `logger.Warnf` is ok 

* if this tool runs "once" but still invoked from shell –log.Fatal could be useful here.

I know we have some potential "fatal" errors like deferred creation of metrics provider. But those can be seen in logs when the tool stops working, and the users can come to logs to investigate (warn is user visible)

grayside: One of the great value adds for this tool would be to include a deployment script that creates alerts on bad behavior from rollouts itself. If my rollout service were to start fatalling a lot, I want push notifications.

cmd/operator/main.go
```diff
@@ -155,11 +157,30 @@ func runDaemon(ctx context.Context, logger *logrus.Logger, cfg *config.Config) e
...
+		var (
+			errs []error
+			wg   sync.WaitGroup
+		)
+		for _, service := range services {
+			wg.Add(1)
+			go func(ctx context.Context, lg *logrus.Logger, svc *rollout.ServiceRecord, strategy *config.Strategy) {
+				err = handleRollout(ctx, lg, svc, strategy)
+				if err != nil {
+					errs = append(errs, err)
```
grayside, line 169: It seems unsafe to simply append to a global slice from inside the go func. I'd expect a RW mutex.

gvso: Oohhh, true. Good catch!

cmd/operator/main.go
ahmetb: I think ideally we would expose that functionality through a configuration file. You can come in and specify which channels to notify (pubsub, slack, pagerduty)...

grayside: Much later, to clarify: I mean bootstrapping release manager's service updates by releasing gradually with release manager :)

ahmetb: haha that would be totally interesting :D though ideally we would have only 1 version of RM running, and ideally that'd be the latest, so that old/new versions aren't racing/conflicting.

PR #42, opened 2020-07-30
config: Add configuration for min request count

gvso:
This adds configuration and flag to specify the minimum number of requests needed to diagnose the candidate's health

### DIFF
--- a/cmd/operator/main.go
+++ b/cmd/operator/main.go
@@ -70,13 +70,14 @@ var (
 	flRegionsString string
 
 	// Rollout strategy-related flags.
-	flSteps       stepFlags
-	flStepsString string
-	flInterval    int64
-	flErrorRate   float64
-	flLatencyP99  float64
-	flLatencyP95  float64
-	flLatencyP50  float64
+	flSteps           stepFlags
+	flStepsString     string
+	flInterval        int64
+	flMinRequestCount uint64
+	flErrorRate       float64
+	flLatencyP99      float64
+	flLatencyP95      float64
+	flLatencyP50      float64
 
 	// Metrics provider flags.
 	flGoogleSheetsID string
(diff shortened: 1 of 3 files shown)


cmd/operator/main.go
```diff
@@ -92,6 +93,7 @@ func init() {
 	flag.Var(&flSteps, "step", "a percentage in traffic the candidate should go through")
 	flag.StringVar(&flStepsString, "steps", "5,20,50,80", "define steps in one flag separated by commas (e.g. 5,30,60)")
 	flag.Int64Var(&flInterval, "interval", 0, "the time between each rollout step")
+	flag.Uint64Var(&flMinRequestCount, "min-requests", 0, "expected minimum requests before determining candidate's health")
```
ahmetb, line 96: Lets set a default value

pkg/config/config.go
```diff
@@ -104,19 +105,22 @@ func (config *Config) Validate(cliMode bool) error {
 
 func validateMetrics(metricsCriteria Metric) error {
 	threshold := metricsCriteria.Threshold
+	if threshold < 0 {
```
ahmetb, line 108: Why not move it inside case: ?

gvso: That would mean checking for `threshold < 0` in 3 parts (more times as we support more metrics). Also, I can't think of a metrics that can have negative thresholds, which might be possible, but we haven't dealt with one yet.

cmd/operator/main.go
gvso: Let me research and think of a potential default value

gvso: I've been thinking about this, but it becomes hard to set a default value that works for many (most) services. This value really depends on the traffic your service often receives, which might require some analysis of the traffic in the past (which is possible). Also, this value also depends on the decision in #43 

If we decide to do #43, it might be easier to go with some "random" number like 1000.

ahmetb: Per #43 I think we should go with default=0 (no such requirement).

PR #44, opened 2020-07-31
health: Add tests for thresholds and metrics at 0

### LINKED ISSUE #31: health: Add tests for metrics and thresholds set at 0
As discussed [here](https://github.com/GoogleCloudPlatform/cloud-run-release-operator/pull/15#discussion_r460335566), it might be good to test metrics and thresholds set to 0

gvso:
This adds two test cases. One test is for all thresholds set at 0, while the second is for all metrics values being 0. This closes #31

### DIFF
--- a/pkg/health/health_test.go
+++ b/pkg/health/health_test.go
@@ -30,16 +30,8 @@ func TestDiagnosis(t *testing.T) {
 			expected: health.Diagnosis{
 				OverallResult: health.Healthy,
 				CheckResults: []health.CheckResult{
-					{
-						Threshold:     750.0,
-						ActualValue:   500.0,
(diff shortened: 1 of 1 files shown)


pkg/health/health_test.go
```diff
@@ -92,17 +72,39 @@ func TestDiagnosis(t *testing.T) {
...
-						Threshold:     0.95,
-						ActualValue:   1.0,
-						IsCriteriaMet: false,
-					},
+					{Threshold: 0.95, ActualValue: 1.0},
 				},
 			},
 		},
 		{
-			name:      "should err, empty health criteria",
```
gvso, line 104: This test is repeated 2 tests below, so I removed this one.

PR #47, opened 2020-07-31
Add basic Dockerfile

gvso:
This is a simple Dockerfile for running the tool in CLI mode. 

To deal with Google Cloud credentials, mounting `gcloud`'s default credentials should be enough
```
gcloud auth application-default login
docker run -v=$HOME/.config/gcloud:/root/.config/gcloud release-operator -project=myproject -interval=600
```


Dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM golang:1.14
```
ahmetb, line 1: use multi stage dockerfiles pls

Dockerfile
```diff
@@ -0,0 +1,9 @@
...
+
+RUN mkdir /src
```
ahmetb, line 3: use WORKDIR

Dockerfile
```diff
@@ -0,0 +1,9 @@
...
+COPY . /src
+WORKDIR /src/cmd/operator
+
+RUN go build -mod=readonly -o operator
```
ahmetb, line 7: add some static compilation flags so it can work without glibc

Dockerfile
gvso: Hmm, distroless images. That's interesting 😃

Dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM golang:1.14 AS builder
+WORKDIR /src
+COPY . .
+RUN CGO_ENABLED=0 GOOS=linux go build -mod=readonly -v -o operator ./cmd/operator
```
ahmetb, line 4: `-o /bin/operator`

good name for binary: `cloud_run_release_operator`

Dockerfile
```diff
@@ -0,0 +1,9 @@
...
+
+FROM gcr.io/distroless/static
+WORKDIR /root
```
ahmetb, line 7: you can just do this `COPY --from=builder /bin/operator /bin/operator`
ENTRYPOINT ["/bin/operator", "-cli"]

Dockerfile
```diff
@@ -0,0 +1,8 @@
+FROM golang:1.14 AS builder
+WORKDIR /src
+COPY . .
```
ahmetb, line 3: to speed up builds (at least locally on dev machine) you can do:

```
COPY go.mod go.sum ./
RUN go mod download (+ whatever readonly flag goes here if any)
COPY . . 
[...]
```

PR #50, opened 2020-08-03
main: Flags for rollout interval and health offset

gvso:
This separates the flags for interval between each rollout process and the health check offset. It sets a default rollout interval of 60 seconds and a default health check of 10 minutes.


gvso:
@ahmetb This addresses the separation of these config values that we discussed at #46 

P.s.: It seems like your review was not submitted in that PR.

pkg/rollout/rollout.go
```diff
@@ -304,8 +304,7 @@ func (r *Rollout) updateAnnotations(svc *run.Service, stable, candidate string)
 
 // diagnoseCandidate returns the candidate's diagnosis based on metrics.
 func (r *Rollout) diagnoseCandidate(candidate string, healthCriteria []config.Metric) (d health.Diagnosis, err error) {
-	// TODO(gvso): Consider using a different config value for the offset.
-	healthCheckOffset := time.Duration(r.strategy.Interval) * time.Second
+	healthCheckOffset := time.Duration(r.strategy.HealthIntervalMinute) * time.Minute
```
ahmetb, line 307: might be wise to accept `int` or `int64` for these types (instead of uint64).

I believe that way we can directly multiply e.g. `:= var * time.Minute`. though needs double-checking.

pkg/config/config.go
```diff
@@ -78,8 +76,8 @@ func NewTarget(project string, regions []string, labelSelector string) *Target {
 
 // Validate checks if the configuration is valid.
 func (config *Config) Validate(cliMode bool) error {
-	if cliMode && config.Strategy.Interval <= 0 {
-		return errors.New("time interval must be greater than 0")
+	if cliMode && config.Strategy.HealthIntervalMinute <= 0 {
```
ahmetb, line 79: it's a uint. :) how can be <= 0?
I recommend just making int

gvso: Just of curiosity. If we're gonna check for `<= 0` we make it an `int`, why is an `int` better than `uint` in this type of contexts? Of course, this check should go away if we used `uint` as you point out.

ahmetb: Mostly because we can validate ints better and we don't make advantage of uint.

cmd/operator/main.go
```diff
@@ -57,29 +57,26 @@ func (steps stepFlags) String() string {
...
-	flInterval        int64
-	flMinRequestCount uint64
-	flErrorRate       float64
-	flLatencyP99      float64
-	flLatencyP95      float64
-	flLatencyP50      float64
+	flSteps              stepFlags
+	flStepsString        string
+	flHealthOffsetMinute int64
+	flMinRequestCount    uint64
```
ahmetb, line 75: make int

cmd/operator/main.go
```diff
@@ -94,12 +91,13 @@ func init() {
...
 	flag.Var(&flSteps, "step", "a percentage in traffic the candidate should go through")
 	flag.StringVar(&flStepsString, "steps", "5,20,50,80", "define steps in one flag separated by commas (e.g. 5,30,60)")
-	flag.Int64Var(&flInterval, "interval", 0, "the time between each rollout step")
+	flag.Int64Var(&flHealthOffsetMinute, "healthcheck-offset", 10, "use metrics from the last N minutes relative to current rollout process")
 	flag.Uint64Var(&flMinRequestCount, "min-requests", 0, "expected minimum requests before determining candidate's health")
 	flag.Float64Var(&flErrorRate, "max-error-rate", 1.0, "expected max server error rate (in percent)")
 	flag.Float64Var(&flLatencyP99, "latency-p99", 0, "expected max latency for 99th percentile of requests (set 0 to ignore)")
 	flag.Float64Var(&flLatencyP95, "latency-p95", 0, "expected max latency for 95th percentile of requests (set 0 to ignore)")
 	flag.Float64Var(&flLatencyP50, "latency-p50", 0, "expected max latency for 50th percentile of requests (set 0 to ignore)")
+	flag.Int64Var(&flCLILoopIntervalSec, "cli-run-interval", 60, "the time between each rollout process (in seconds)")
```
ahmetb, line 100: next to the -cli flag (both in variable list and in -help)

pkg/config/config.go
```diff
@@ -78,8 +76,8 @@ func NewTarget(project string, regions []string, labelSelector string) *Target {
 
 // Validate checks if the configuration is valid.
 func (config *Config) Validate(cliMode bool) error {
-	if cliMode && config.Strategy.Interval <= 0 {
-		return errors.New("time interval must be greater than 0")
+	if cliMode && config.Strategy.HealthOffsetMinute <= 0 {
+		return errors.New("health check offset must be positive")
```
ahmetb, line 80: errorf

pkg/rollout/rollout.go
```diff
@@ -304,8 +304,11 @@ func (r *Rollout) updateAnnotations(svc *run.Service, stable, candidate string)
 
 // diagnoseCandidate returns the candidate's diagnosis based on metrics.
 func (r *Rollout) diagnoseCandidate(candidate string, healthCriteria []config.Metric) (d health.Diagnosis, err error) {
-	// TODO(gvso): Consider using a different config value for the offset.
-	healthCheckOffset := time.Duration(r.strategy.Interval) * time.Second
+	if r.strategy.HealthOffsetMinute <= 0 {
+		return d, errors.New("health check offset must be positive")
+	}
```
ahmetb, line 309: too late to check for this

PR #51, opened 2020-08-04
rollout: Include health report in service object

gvso:
This adds annotation about the last health report. It uses the key rollout.cloud.run/lastHealthReport and assigns it a string with the last diagnosis (healthy, unhealthy) and the check results.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -6,7 +6,6 @@ require (
 	cloud.google.com/go v0.60.0 // indirect
 	github.com/TV4/logrus-stackdriver-formatter v0.1.0
 	github.com/go-stack/stack v1.8.0 // indirect
-	github.com/google/go-cmp v0.5.0
 	github.com/mattn/go-isatty v0.0.12
 	github.com/pkg/errors v0.9.1
 	github.com/sirupsen/logrus v1.6.0
--- a/pkg/health/health.go
+++ b/pkg/health/health.go
@@ -22,6 +22,19 @@ const (
 	Unhealthy
 )
 
+func (d DiagnosisResult) String() string {
+	switch d {
+	case Inconclusive:
+		return "inconclusive"
+	case Healthy:
+		return "healthy"
+	case Unhealthy:
+		return "unhealthy"
+	default:
+		return "unknown"
+	}
+}
+
 // Diagnosis is the information about the health of the revision.
 type Diagnosis struct {
 	OverallResult DiagnosisResult
--- a/pkg/health/report.go
+++ b/pkg/health/report.go
@@ -8,10 +8,9 @@ import (
 	"github.com/pkg/errors"
 )
 
-// StringReport report generates a human-readable report of the diagnosis.
+// StringReport returns a human-readable report of the diagnosis.
 //
 // The returned string has the format:
-//
 // status: healthy
 // metrics:
 // - request-latency[p99]: 500.00 (threshold 750.0)
--- a/pkg/health/report.go
+++ b/pkg/health/report.go
@@ -1,0 +1,37 @@
+package health
+
+import (
+	"encoding/json"
+
+	"github.com/GoogleCloudPlatform/cloud-run-release-operator/pkg/config"
+	"github.com/pkg/errors"
+)
+
+// JSONReport returns a JSON representation of the diagnosis.
+func JSONReport(healthCriteria []config.Metric, diagnosis Diagnosis) (string, error) {
+	var resultsMap []map[string]interface{}
+	for i, result := range diagnosis.CheckResults {
+		criteria := healthCriteria[i]
+		report := map[string]interface{}{
+			"metricsType":   criteria.Type,
+			"threshold":     result.Threshold,
+			"actualValue":   result.ActualValue,
+			"isCriteriaMet": result.IsCriteriaMet,
+		}
+		if criteria.Type == config.LatencyMetricsCheck {
+			report["percentile"] = criteria.Percentile
+		}
+		resultsMap = append(resultsMap, report)
+	}
+
+	report := struct {
+		Diagnosis    string                   `json:"diagnosis"`
+		CheckResults []map[string]interface{} `json:"checkResults"`
+	}{
+		Diagnosis:    diagnosis.OverallResult.String(),
+		CheckResults: resultsMap,
+	}
+
+	reportJSON, err := json.Marshal(report)
+	return string(reportJSON), errors.Wrap(err, "failed to marshal report")
+}
--- a/pkg/health/report.go
+++ b/pkg/health/report.go
@@ -40,6 +40,9 @@ func StringReport(healthCriteria []config.Metric, diagnosis Diagnosis) string {
 }
 
 // JSONReport returns a JSON representation of the diagnosis.
+//
+// TODO: consider if this function is useful (e.g. for a dashboard) since it's
+// not used anywhere yet.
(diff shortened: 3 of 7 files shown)


go.mod
```diff
@@ -6,7 +6,6 @@ require (
 	cloud.google.com/go v0.60.0 // indirect
 	github.com/TV4/logrus-stackdriver-formatter v0.1.0
 	github.com/go-stack/stack v1.8.0 // indirect
-	github.com/google/go-cmp v0.5.0
```
gvso, line 9: testify is able to compare more than primitives objects, so I really didn't need go-cmp before.

pkg/health/report.go
```diff
@@ -0,0 +1,37 @@
...
+		criteria := healthCriteria[i]
+		report := map[string]interface{}{
+			"metricsType":   criteria.Type,
+			"threshold":     result.Threshold,
+			"actualValue":   result.ActualValue,
+			"isCriteriaMet": result.IsCriteriaMet,
+		}
+		if criteria.Type == config.LatencyMetricsCheck {
+			report["percentile"] = criteria.Percentile
+		}
```
gvso, line 23: I also considered using the `health.Diagnosis` struct directly, but there were two drawbacks
1) We have previously decided not to include the metrics type in `health.CheckResult` since that would be repeating the same data. Plus, the percentile is also not part of that struct.
2) The `OverallDiagnosis` property is a `DiagnosisResult` (`int), so the JSON representation turns out to be a value 0..3, which is not that helpful for our users.

gvso:
Btw, in the project document, we talk about storing a history of *all* decisions. This makes this annotation seem redundant, but it might be useful to find the know why the last decision was taken faster. 

If not, we can remove the `lastHealthReport` annotation part and keep the logic to generate the an element of the history of decisions.

pkg/health/health.go
```diff
@@ -22,6 +22,19 @@ const (
 	Unhealthy
 )
 
+func (d DiagnosisResult) String() string {
```
ahmetb, line 25: +++ good way

pkg/health/report.go
```diff
@@ -0,0 +1,69 @@
...
+		if criteria.Type == config.LatencyMetricsCheck {
+			report += fmt.Sprintf("\n- %s[p%.0f]: %.2f (threshold %.2f)", criteria.Type, criteria.Percentile, result.ActualValue, criteria.Threshold)
+			continue
+		}
+
+		format := "\n- %s: %.2f (threshold %.2f)"
+		if criteria.Type == config.RequestCountMetricsCheck {
+			// No decimals for request count.
+			format = "\n- %s: %.0f (threshold %.0f)"
+		}
```
gvso, line 35: Hopefully, this is readable enough and the example in the comment make it clear. If not, any recommendations?

pkg/health/report.go
```diff
@@ -0,0 +1,72 @@
...
+		report += fmt.Sprintf(format, criteria.Type, result.ActualValue, criteria.Threshold)
+	}
+
+	return report
+}
+
+// JSONReport returns a JSON representation of the diagnosis.
+//
+// TODO: consider if this function is useful (e.g. for a dashboard) since it's
+// not used anywhere yet.
```
gvso, line 45: I kept this function since we might need it eventually and the tests are a little bit "tedious" to write.

pkg/health/report.go
```diff
@@ -0,0 +1,71 @@
...
+func JSONReport(healthCriteria []config.Metric, diagnosis Diagnosis) (string, error) {
```
ahmetb, line 45: I'd delete json for now

gvso: I deleted it. I am keeping this branch in my fork, so I have access to that work anyways.

pkg/health/report.go
```diff
@@ -0,0 +1,37 @@
...
+	"github.com/GoogleCloudPlatform/cloud-run-release-operator/pkg/config"
+)
+
+// StringReport returns a human-readable report of the diagnosis.
+//
+// The returned string has the format:
+// status: healthy
+// metrics:
+// - request-latency[p99]: 500.00 (needs 750.0)
+// - request-count: 800 (needs 1000)
```
ahmetb, line 15: this will most likely go out of date, I'd refrain from listing it here, but it's ok

pkg/health/report.go
```diff
@@ -0,0 +1,37 @@
...
+func StringReport(healthCriteria []config.Metric, diagnosis Diagnosis) string {
+	report := fmt.Sprintf("status: %s\n", diagnosis.OverallResult.String())
```
ahmetb, line 17: what if we also added the time we wrote this down?
does the user have any way of telling when was this status report for?

pkg/rollout/rollout.go
```diff
@@ -114,7 +122,7 @@ func (r *Rollout) UpdateService(svc *run.Service) (*run.Service, error) {
 	if isNewCandidate(svc, candidate) {
 		r.log.Debug("new candidate, assign some traffic")
 		svc = r.PrepareRollForward(svc, stable, candidate)
-		svc = r.updateAnnotations(svc, stable, candidate)
+		svc = r.updateAnnotations(svc, stable, candidate, "new candidate, no health report available yet")
```
ahmetb, line 125: this is not a great way. I'll offer alternatives.

1.  pass a `statusMsgFunc func()string` that returns the msg you wanna set.
2. create a new method `svc = r.setAnnotation(key, msg)`

pkg/health/report.go
gvso: We could, but I'd probably have a function `setHealthReportAnnotation(report string)` in `rollout` since there are cases in which this function is not called (e.g. new candidate). In that function, we can append the time at the end of any report.

I will defer this after #57 since we would need the mocked time introduced there. Doing it here would be double work and would make this PR much longer.

pkg/rollout/rollout.go
```diff
@@ -286,28 +293,31 @@ func (r *Rollout) nextCandidateTraffic(current int64) int64 {
...
 	if r.promoteToStable {
-		svc.Metadata.Annotations[StableRevisionAnnotation] = candidate
+		setAnnotation(svc, StableRevisionAnnotation, candidate)
 		delete(svc.Metadata.Annotations, CandidateRevisionAnnotation)
 		return svc
 	}
 
-	svc.Metadata.Annotations[StableRevisionAnnotation] = stable
-	svc.Metadata.Annotations[CandidateRevisionAnnotation] = candidate
+	setAnnotation(svc, StableRevisionAnnotation, stable)
```
ahmetb, line 304: nice, yeah you can just use it here :D

pkg/rollout/rollout.go
```diff
@@ -122,7 +122,12 @@ func (r *Rollout) UpdateService(svc *run.Service) (*run.Service, error) {
 	if isNewCandidate(svc, candidate) {
 		r.log.Debug("new candidate, assign some traffic")
 		svc = r.PrepareRollForward(svc, stable, candidate)
-		svc = r.updateAnnotations(svc, stable, candidate, "new candidate, no health report available yet")
+		svc = r.updateAnnotations(svc, stable, candidate)
+
+		// TODO(gvso): create a setHealthReportAnnotation that appends the
+		// current time to the report before setting the annotation.
```
ahmetb, line 128: this probably could be part of the health report IMO

gvso: My initial thought was also this one, but the problem is that we don't generate a report when we have a new candidate. We have a string literal in that case

PR #52, opened 2020-08-04
main: Support rollouts through HTTP requests

gvso:
This creates an HTTP server when a port is specified using `-http-addr`. It also separates the rollout logic for reuse.

### DIFF
--- a/cmd/operator/main.go
+++ b/cmd/operator/main.go
@@ -58,7 +58,7 @@ var (
 	flLoggingLevel       string
 	flCLI                bool
 	flCLILoopIntervalSec int
-	flPort               int
+	flHTTPAddr           string
 	flProject            string
 	flLabelSelector      string
 
@@ -81,19 +81,15 @@ var (
 )
 
 func init() {
-	defaultPort := 8080
+	defaultAddr := ":8080"
 	if v := os.Getenv("PORT"); v != "" {
-		var err error
-		defaultPort, err = strconv.Atoi(v)
-		if err != nil {
-			logrus.WithField("v", v).Fatalf("invalid PORT environment variable, expected int")
-		}
+		defaultAddr = fmt.Sprintf(":%s", v)
 	}
 
 	flag.StringVar(&flLoggingLevel, "verbosity", "info", "the logging level (e.g. debug)")
 	flag.BoolVar(&flCLI, "cli", false, "run as CLI application to manage rollout in intervals")
 	flag.IntVar(&flCLILoopIntervalSec, "cli-run-interval", 60, "the time between each rollout process (in seconds)")
-	flag.IntVar(&flPort, "-port", defaultPort, "port where to listen to http requests (e.g. 8080)")
+	flag.StringVar(&flHTTPAddr, "http-addr", defaultAddr, "port where to listen to http requests (e.g. 8080)")
 	flag.StringVar(&flProject, "project", "", "project in which the service is deployed")
 	flag.StringVar(&flLabelSelector, "label", "rollout-strategy=gradual", "filter services based on a label (e.g. team=backend)")
 	flag.StringVar(&flRegionsString, "regions", "", "the Cloud Run regions where the service should be looked at")
@@ -150,9 +146,8 @@ func main() {
 		runDaemon(ctx, logger, cfg)
 	} else {
 		http.HandleFunc("/rollout", makeRolloutHandler(logger, cfg))
-		logger.WithField("port", flPort).Infof("starting server")
-		port := fmt.Sprintf(":%d", flPort)
-		logger.Fatal(http.ListenAndServe(port, nil))
+		logger.WithField("addr", flHTTPAddr).Infof("starting server")
+		logger.Fatal(http.ListenAndServe(flHTTPAddr, nil))
 	}
 }
 
--- a/cmd/operator/main.go
+++ b/cmd/operator/main.go
@@ -143,7 +143,8 @@ func main() {
 		}
 	}
 
-	http.HandleFunc("/", makeRolloutHandler(logger, cfg))
+	http.HandleFunc("/rollout", makeRolloutHandler(logger, cfg))
+	logger.Infof("starting server at %s", flHTTPAddr)
 	logger.Fatal(http.ListenAndServe(flHTTPAddr, nil))
 }
 
--- a/cmd/operator/main.go
+++ b/cmd/operator/main.go
@@ -17,6 +17,7 @@ package main
 import (
 	"context"
 	"flag"
+	"fmt"
 	"net/http"
 	"os"
 	"strconv"
@@ -57,7 +58,7 @@ var (
 	flLoggingLevel       string
 	flCLI                bool
 	flCLILoopIntervalSec int
-	flHTTPAddr           string
+	flPort               int
 	flProject            string
 	flLabelSelector      string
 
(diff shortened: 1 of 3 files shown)


cmd/operator/main.go
```diff
@@ -147,6 +144,10 @@ func main() {
 			logger.Fatalf("error when running daemon: %v", err)
 		}
 	}
+
+	http.HandleFunc("/rollout", makeRolloutHandler(logger, cfg))
```
ahmetb, line 148: let's change the -http-addr flag a bit.

1. edit its help message, it looks odd

2. right now default is "", you can give:

```
defaultAddr == ""
if v:= os.Getenv("PORT"); v!="" {
  defaultAddr = ":8080"
}
```

cmd/operator/rollout.go
```diff
@@ -0,0 +1,78 @@
...
+	for _, svc := range svcs {
+		wg.Add(1)
+		go func(ctx context.Context, lg *logrus.Logger, svc *rollout.ServiceRecord, strategy *config.Strategy) {
+			err := handleRollout(ctx, lg, svc, strategy)
+			if err != nil {
+				mu.Lock()
+				errs = append(errs, err)
+				mu.Unlock()
+			}
+			wg.Done()
```
ahmetb, line 31: let's move to the beginning of the function + defer

cmd/operator/server.go
```diff
@@ -0,0 +1,37 @@
...
+		services, err := getTargetedServices(ctx, logger, cfg.Targets)
+		if err != nil {
+			logger.Errorf("failed to get targeted services: %v", err)
+			w.WriteHeader(http.StatusInternalServerError)
+			fmt.Fprintf(w, "failed to retrieved services %v", err)
+			return
+		}
+		if len(services) == 0 {
+			logger.Warn("no service matches the targets")
+		}
```
ahmetb, line 24: move this into runRollouts because it's (1) duplicated (2) not handled differently than any error right now

cmd/operator/server.go
```diff
@@ -0,0 +1,37 @@
...
+
+		// TODO(gvso): Filter "fatal" errors from "no-bad" errors (e.g. no
+		// requests in interval when getting metrics).
+		errs := runRollouts(ctx, logger, services, cfg.Strategy)
+		errsStr := rolloutErrsToString(errs)
+		if len(errs) != 0 {
+			msg := fmt.Sprintf("there were %d errors: %s", len(errs), errsStr)
+			logger.Warn(msg)
```
ahmetb, line 32: this is great

cmd/operator/server.go
```diff
@@ -0,0 +1,37 @@
...
+			w.WriteHeader(http.StatusInternalServerError)
```
ahmetb, line 33: since we'll be called from Cloud Scheduler, let's not cause retries.

cmd/operator/main.go
gvso: We actually need to check for more things here, since there are 3 scenarios (in order of precedence):

1. The `PORT` variable is in the environment
2. The `-http-addr` flag
3. The default ":8080"

Because there's a default and the env variable has precedence, I think we need to drop the checks for just one of `-cli` or `-http-addr`. So, by default, we can assume the developer is running the tool as a webserver listening on `:8080` unless you specify you want to run it as CLI (`-cli` flag).

gvso: I understand the point of doing 2, but I think we can determine the port later in the program instead of passing a default to the flag based on a env variable. If we do 2, the user might become confuse when they run

```
./operator -cli ...
```
and get the error "only one of -cli or -http-addr can be used" just because `PORT` was set in their env variables.

ahmetb: 🤔 yeah I see. In that case, what if we only relied on `-cli` to determine cli mode or not?
alternatively we can introduce `-server` option, which is mutually exclusive with `-cli`, this seems cleaner.

cmd/operator/rollout.go
gvso: I agree with this, but I just realized that if we do this, `runDaemon` [can't ever return an error](https://github.com/GoogleCloudPlatform/cloud-run-release-operator/blob/c6b0627606cca06a83a0828fecce8b5042f3cadf/cmd/operator/main.go#L156). 

But this is not a big problem since getting services mostly fail either because 
1. a Cloud Run region is down (like last week) 
2. wrong region is specified (e.g. `us-east`, notice typo)

For 1, we can make the argument that's not the tool's problem and keeping it running is actually the right decision. For 2, the user should quickly realized of the problem since it'd happen for the first time right at the beginning of the execution.

cmd/operator/server.go
gvso: I agree with this, but I just realized that if we do this, `runDaemon` [can't ever return an error](https://github.com/GoogleCloudPlatform/cloud-run-release-operator/blob/c6b0627606cca06a83a0828fecce8b5042f3cadf/cmd/operator/main.go#L156). 

But this is not a big problem since getting services mostly fail either because 
1. a Cloud Run region is down (like last week) 
2. wrong region is specified (e.g. `us-east`, notice typo)

For 1, we can make the argument that it's not the tool's problem and keeping it running is actually the right decision. For 2, the user should quickly realize of the problem since it'd happen for the first time right at the beginning of the execution.

cmd/operator/main.go
gvso: I renamed it to `port` expecting an `int`. We're getting the default from the environment while making the HTTP the default mode. So, `-cli` should be used to run it in the terminal.

ahmetb: I don't recommend that. we still need to have a way of specifying addr e.g. `localhost:8080` vs `:8080` are different things. 

Also we should still use $PORT to use `:$PORT` if `-http-addr` is unspecified (e.g. on Cloud Run )

gvso: Then, we can relied only `-cli` as you pointed out and make the HTTP server the default one with default addr `:8080` which can be overwritten by `$PORT`. However, if `-http-addr` is specified, that has the highest precedence.

cmd/operator/main.go
```diff
@@ -83,10 +81,15 @@ var (
...
+	if v := os.Getenv("PORT"); v != "" {
+		defaultAddr = fmt.Sprintf(":%s", v)
+	}
+
 	flag.StringVar(&flLoggingLevel, "verbosity", "info", "the logging level (e.g. debug)")
 	flag.BoolVar(&flCLI, "cli", false, "run as CLI application to manage rollout in intervals")
 	flag.IntVar(&flCLILoopIntervalSec, "cli-run-interval", 60, "the time between each rollout process (in seconds)")
-	flag.StringVar(&flHTTPAddr, "http-addr", "", "listen on http portrun on request (e.g. :8080)")
+	flag.StringVar(&flHTTPAddr, "http-addr", defaultAddr, "port where to listen to http requests (e.g. 8080)")
 	flag.StringVar(&flProject, "project", "", "project in which the service is deployed")
```
ahmetb, line 93: on user's computer we can define project as `gcloud config get-value core/project -q` + trim output. 

on Cloud Run (by looking at $K_SERVICE) https://github.com/GoogleCloudPlatform/cloud-run-hello/blob/9b0e68f416d5f3f23398bad81d57f1e268a99ad8/hello.go#L37-L52

if you detect something successfully: log it user-visible.

cmd/operator/main.go
```diff
@@ -83,10 +81,15 @@ var (
...
+	defaultAddr := ":8080"
+	if v := os.Getenv("PORT"); v != "" {
+		defaultAddr = fmt.Sprintf(":%s", v)
+	}
+
 	flag.StringVar(&flLoggingLevel, "verbosity", "info", "the logging level (e.g. debug)")
 	flag.BoolVar(&flCLI, "cli", false, "run as CLI application to manage rollout in intervals")
 	flag.IntVar(&flCLILoopIntervalSec, "cli-run-interval", 60, "the time between each rollout process (in seconds)")
-	flag.StringVar(&flHTTPAddr, "http-addr", "", "listen on http portrun on request (e.g. :8080)")
+	flag.StringVar(&flHTTPAddr, "http-addr", defaultAddr, "port where to listen to http requests (e.g. 8080)")
```
ahmetb, line 92: add : back to e.g.

cmd/operator/rollout.go
```diff
@@ -0,0 +1,86 @@
...
+		lg.Debug("service kept unchanged")
+	}
+	return nil
+}
+
+// rolloutErrsToString returns the string representation of all the errors found
+// during the rollout of all targeted services.
+func rolloutErrsToString(errs []error) (errsStr string) {
+	for i, err := range errs {
+		errsStr += fmt.Sprintf("\n[error#%d] %v", i, err)
```
ahmetb, line 83: this always starts with \n
prefix it with `multiple errors: `

cmd/operator/rollout.go
```diff
@@ -0,0 +1,86 @@
...
+		mu   sync.Mutex
+		wg   sync.WaitGroup
+	)
+	for _, svc := range svcs {
+		wg.Add(1)
+		go func(ctx context.Context, lg *logrus.Logger, svc *rollout.ServiceRecord, strategy *config.Strategy) {
+			defer wg.Done()
+			err := handleRollout(ctx, lg, svc, strategy)
+			if err != nil {
+				mu.Lock()
```
ahmetb, line 36: let's add

log.Debugf("rollout error for service: %+v")

Issue #53, opened 2020-08-05
Add a minimum time between each step


gvso:
In #50, we're introducing a flag to check for metrics in the last N minutes while we are expecting to execute every rollout process after an interval (`-cli`) or periodically using HTTP requests. The issue with this approach is that if the interval between each rollout process and the health check offset is significant, the candidate can become stable very quickly even if buggy. This is because the health checks of the previous rollout process will have a very big weight when determining the candidate's health

**Example:**
Interval between each rollout process: 5 minutes
Health check offset: 30 minutes

![Figure](https://user-images.githubusercontent.com/2955158/89429145-30b4f880-d70b-11ea-9479-92eb4bef27e7.png)

The health check at min 35 might determine a healthy candidate even though candidate is buggy because most of the metrics is based on the previous 25 minutes before the current step (traffic percentage).

To solve this, an alternative is to allow the user to define the minimum time between each step (increment in traffic percentage). So, if the metrics show a healthy candidate but no enough time has passed, we do not roll forward. However, if the metrics show an unhealthy candidate, we rollback independent of the time between the current and last rollout process.

Issue #43, opened 2020-07-31
Make health check offset based on last updated time


gvso:
In #38, we introduced rollbacks and integration with the `health` package. When obtaining the metrics, we passed an offset of how much time in the past we want to look up (e.g. 10 minutes). 

The current implementation always look back at a constant number of time (e.g. 10 minutes, 1 hour, etc.). This has the (potential) advantage that it allows us to measure the latency/error rate given an average load time (e.g. `6000 requests in 10 minutes = 6 request/s`). However, it also has the disadvantage of potentially taking longer to completely rollout a new revision (or not rolling out at all if an unreachable request count is set). For instance, if the min request count is 6000 and we check with a constant offset of 10 minutes, it can happen that the new revision always get a number of requests that is below 6000 in the last 10 minutes. This is especially more likely when the candidate gets small share of the traffic (at the beginning of the rollout).

The alternative solution would be to add an annotation about the last time the candidate's traffic was increased, so we can calculate an `offset = time.Now() - lastTime`

gvso:
@ahmetb @grayside Any thoughts on this?

ahmetb:
I think the "we look at last N minutes" approach is easier to wrap your mind around (for users) and easier for us to implement (no need to keep track of last checked date).

For crowded services, setting a decent `<min request, last N minutes>` pair should be fairly easy for the operator of the service.

After all the point of "min requests" criterion is to ensure there were at least some requests and the monitoring signals aren't anecdotal.

Maybe ship with "default min requests=0" (i.e. no such requirement).

PR #56, opened 2020-08-05
main: Add flag for min time before roll forward

gvso:


cmd/operator/main.go
```diff
@@ -96,6 +97,7 @@ func init() {
 	flag.Var(&flSteps, "step", "a percentage in traffic the candidate should go through")
 	flag.StringVar(&flStepsString, "steps", "5,20,50,80", "define steps in one flag separated by commas (e.g. 5,30,60)")
 	flag.IntVar(&flHealthOffsetMinute, "healthcheck-offset", 30, "use metrics from the last N minutes relative to current rollout process")
+	flag.IntVar(&flMinForwardTimeMinute, "min-forward-time", 30, "the minimum time needed to roll the candidate forward (in minutes)")
```
ahmetb, line 100: - `min-wait` is a better name
- desc: minimum time to wait between rollout stages (in minutes) use 0 to disable

pkg/config/config.go
```diff
@@ -39,9 +39,10 @@ type Metric struct {
 
 // Strategy is the steps and configuration for rollout.
 type Strategy struct {
-	Steps              []int64
-	HealthCriteria     []Metric
-	HealthOffsetMinute int
+	Steps                 []int64
+	HealthCriteria        []Metric
+	HealthOffsetMinute    int
+	TimeRollForwardMinute int
```
ahmetb, line 45: var name isn't very descriptive.

TimeBetweenRolloutsMins is better?

pkg/config/config.go
```diff
@@ -54,13 +55,14 @@ type Config struct {
 }
 
 // WithValues initializes a configuration with the given values.
-func WithValues(targets []*Target, steps []int64, healthOffset int, metrics []Metric) *Config {
+func WithValues(targets []*Target, steps []int64, healthOffset, timeForward int, metrics []Metric) *Config {
```
ahmetb, line 58: again var name isn't good

cmd/operator/main.go
gvso: Wouldn't `min-wait` be confused with `cli-run-interval`? Both are "minimum times to wait"?

ahmetb: paging @grayside for an extra pair of eyes :D

gvso: @grayside For context of where this is coming from, see #53

pkg/config/config.go
gvso: I renamed it to `TimeBetweenRollForwardsMins`

pkg/config/config_test.go
```diff
@@ -18,58 +18,58 @@ func TestIsValid(t *testing.T) {
 			name: "correct config with label selector",
 			config: config.WithValues([]*config.Target{
 				config.NewTarget("myproject", []string{"us-east1", "us-west1"}, "team=backend"),
-			}, []int64{5, 30, 60}, 20, nil),
+			}, []int64{5, 30, 60}, 20, 30, nil),
```
grayside, line 21: The list of values in these tests is getting a bit long to exclude keys.

pkg/config/config.go
grayside: Does this not affect rollback timing?

ahmetb: It won't affect rollouts. First failed health check causes rollback (even if min request count criteria is not met).

cmd/operator/main.go
```diff
@@ -96,6 +97,7 @@ func init() {
 	flag.Var(&flSteps, "step", "a percentage in traffic the candidate should go through")
 	flag.StringVar(&flStepsString, "steps", "5,20,50,80", "define steps in one flag separated by commas (e.g. 5,30,60)")
 	flag.IntVar(&flHealthOffsetMinute, "healthcheck-offset", 30, "use metrics from the last N minutes relative to current rollout process")
+	flag.IntVar(&flTimeBeweenRollForwards, "min-wait", 30, "minimum time to wait between rollout stages (in minutes), use 0 to disable")
```
ahmetb, line 100: flag.Duration https://godoc.org/flag#Duration

cmd/operator/main.go
grayside: To restate:
* `--cli-run-interval` a time to wait between triggering the whole set of rollout analysis and operations
* `--min-wait` is proposed for how long to wait after the 1st rollforward of a revision to ensure sufficient metrics are collected to make a decision.

Based on that set of considerations I think I'd make the scope of this more specific, and the "action" use the same terminology to help new users quickly disambiguate the flags.

For example `--monitoring-interval` conveys the point that monitoring is happening over a span of time. Was tempted to bring in the word `evaluation` or `window` as well.

pkg/config/config_test.go
grayside, line 21: This test inputs are getting a bit long to exclude keys.

cmd/operator/main.go
gvso: > --min-wait is proposed for how long to wait after the 1st rollforward of a revision to ensure sufficient metrics are collected to make a decision.
It's actually the time between each rollforward. Sorry if #53 wasn't very clear about this.

> For example --monitoring-interval conveys the point...
Actually, there's another flag `-healthcheck-offset` which basically indicates that metrics from the last `N` minutes should be considered to determine the health.

This flag is different from that idea and `--monitoring-interval` seems similar to `--healthcheck-offset`

pkg/config/config_test.go
gvso: I will address this in a different PR. Keeping track at #59

cmd/operator/main.go
grayside: Interesting, thanks for pointing that out.

In that case I suppose I should ask: What's the use case for this setting to be shorter or longer than the `--healthcheck-offset`? Might this need actually be a "strict" mode?

gvso: There are two reasons for having `healthcheck-offset` different from this "baking" flag:
1) It's easier to implement and allows checking for health given a request load as disccussed in #43 
2) Separating these two things, allow us to do rollbacks as soon as the metrics show signs of failure, but we don't want to do immediate roll forwards because of #53.

gvso: I'm merging this for now. We can address the flag name later. Please let me know if you have any ideas/questions

ahmetb: I am liking the "bake time" notion. `-min-bake-time` and `-health-interval`/`-stability-window` could work. We can address later.

PR #57, opened 2020-08-06
rollout: Roll forward only if enough time passed

gvso:
This adds a new annotation containing the last time the traffic to the candidate was increased. Using the configuration for minimum time between roll forwards, it decides if it can continue rolling the candidate forward.

### DIFF
--- a/pkg/rollout/rollout.go
+++ b/pkg/rollout/rollout.go
@@ -343,7 +343,6 @@ func (r *Rollout) updateAnnotations(svc *run.Service, stable, candidate string) 
 	if r.promoteToStable {
 		setAnnotation(svc, StableRevisionAnnotation, candidate)
 		delete(svc.Metadata.Annotations, CandidateRevisionAnnotation)
-		svc.Metadata.Annotations[LastRolloutAnnotation] = now
 		return svc
 	}
 
--- a/pkg/rollout/rollout.go
+++ b/pkg/rollout/rollout.go
@@ -10,6 +10,7 @@ import (
 	"github.com/GoogleCloudPlatform/cloud-run-release-operator/internal/util"
 	"github.com/GoogleCloudPlatform/cloud-run-release-operator/pkg/config"
 	"github.com/GoogleCloudPlatform/cloud-run-release-operator/pkg/health"
+	"github.com/jonboulle/clockwork"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 	"google.golang.org/api/run/v1"
@@ -20,6 +21,7 @@ const (
 	StableRevisionAnnotation              = "rollout.cloud.run/stableRevision"
 	CandidateRevisionAnnotation           = "rollout.cloud.run/candidateRevision"
 	LastFailedCandidateRevisionAnnotation = "rollout.cloud.run/lastFailedCandidateRevision"
+	LastRolloutAnnotation                 = "rollout.cloud.run/lastRollout"
 	LastHealthReportAnnotation            = "rollout.cloud.run/lastHealthReport"
 )
 
@@ -41,6 +43,7 @@ type Rollout struct {
 	strategy        *config.Strategy
 	runClient       runapi.Client
 	log             *logrus.Entry
+	time            clockwork.Clock
 
 	// Used to determine if candidate should become stable during update.
 	promoteToStable bool
@@ -70,6 +73,7 @@ func New(ctx context.Context, metricsProvider metrics.Provider, svcRecord *Servi
 		region:          svcRecord.Region,
 		strategy:        strategy,
 		log:             logrus.NewEntry(logrus.New()),
+		time:            clockwork.NewRealClock(),
 	}
 }
 
@@ -82,6 +86,12 @@ func (r *Rollout) WithClient(client runapi.Client) *Rollout {
 // WithLogger updates the logger in the rollout instance.
 func (r *Rollout) WithLogger(logger *logrus.Logger) *Rollout {
 	r.log = logger.WithField("project", r.project)
+	return r
+}
+
+// WithClock updates the clock in the rollout instance.
+func (r *Rollout) WithClock(clock clockwork.Clock) *Rollout {
+	r.time = clock
 	return r
 }
 
@@ -138,19 +148,15 @@ func (r *Rollout) UpdateService(svc *run.Service) (*run.Service, error) {
 		return nil, errors.Wrapf(err, "failed to diagnose health for candidate %q", candidate)
 	}
 
-	switch diagnosis.OverallResult {
-	case health.Inconclusive:
-		r.log.Debug("health check inconclusive")
+	svc, err = r.updateServiceBasedOnDiagnosis(svc, diagnosis.OverallResult, stable, candidate)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to update service after diagnosis")
+	}
+	if svc == nil {
+		// If service was unchanged, nil is returned.
+		// TODO(gvso): This should go away once we start getting traffic config
+		// object from updateServiceBasedOnDiagnosis.
 		return nil, nil
-	case health.Healthy:
-		r.log.Debug("healthy candidate, roll forward")
-		svc = r.PrepareRollForward(svc, stable, candidate)
-	case health.Unhealthy:
-		r.log.Info("unhealthy candidate, rollback")
-		r.shouldRollback = true
-		svc = r.PrepareRollback(svc, stable, candidate)
-	default:
-		return nil, errors.Errorf("invalid candidate's health diagnosis %v", diagnosis.OverallResult)
 	}
 
 	svc = r.updateAnnotations(svc, stable, candidate)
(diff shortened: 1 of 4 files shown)


cmd/operator/main.go
```diff
@@ -67,14 +67,15 @@ var (
 	flRegionsString string
```
gvso, line 67: Ignore, same as #56

pkg/config/config.go
```diff
@@ -39,9 +39,10 @@ type Metric struct {
```
gvso, line 39: Ignore, same as #56

pkg/config/config_test.go
```diff
@@ -18,58 +18,58 @@ func TestIsValid(t *testing.T) {
 			name: "correct config with label selector",
```
gvso, line 18: Ignore, same as #56

pkg/rollout/rollout.go
```diff
@@ -200,6 +208,41 @@ func (r *Rollout) PrepareRollback(svc *run.Service, stable, candidate string) *r
 	return svc
 }
 
+// updateServiceBasedOnDiagnosis updates a service's traffic configuration
+// based on the diagnosis.
+// If service was unchanged, nil is returned.
+//
+// TODO (gvso): Refactor this and other methods to return only a traffic object.
+// Then, rename this to trafficBasedOnDiagnosis.
+// Also, move those traffic-config-related methods to traffic.go file.
```
gvso, line 217: Instead of updating the service in `PrepareRollForward` and `PrepareRollForward`, which are down in the call chains, we can just return the traffic config and make the update in, yeah, `UpdateService`.

We need to change the first two method's name and move them (an their helpers) to a new file since this one is getting very big

pkg/config/config.go
```diff
@@ -54,13 +55,14 @@ type Config struct {
 }
 
 // WithValues initializes a configuration with the given values.
-func WithValues(targets []*Target, steps []int64, healthOffset int, metrics []Metric) *Config {
+func WithValues(targets []*Target, steps []int64, healthOffset, timeBetweenRollForwards int, metrics []Metric) *Config {
```
ahmetb, line 58: this should be time.Duration

pkg/rollout/rollout.go
```diff
@@ -10,11 +10,20 @@ import (
...
 	"github.com/sirupsen/logrus"
 	"google.golang.org/api/run/v1"
 )
 
+// Annotations name for information related to the rollout.
+const (
+	StableRevisionAnnotation              = "rollout.cloud.run/stableRevision"
+	CandidateRevisionAnnotation           = "rollout.cloud.run/candidateRevision"
+	LastFailedCandidateRevisionAnnotation = "rollout.cloud.run/lastFailedCandidateRevision"
+	LastRollForwardAnnotation             = "rollout.cloud.run/lastRollForward"
```
ahmetb, line 24: lastRollout

pkg/rollout/rollout.go
```diff
@@ -125,21 +142,12 @@ func (r *Rollout) UpdateService(svc *run.Service) (*run.Service, error) {
...
 	}
 
-	switch diagnosis.OverallResult {
-	case health.Inconclusive:
-		r.log.Debug("health check inconclusive")
+	svc, err = r.updateServiceBasedOnDiagnosis(svc, diagnosis.OverallResult, stable, candidate)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to update service after diagnosis")
+	}
+	if svc == nil {
```
ahmetb, line 149: add a comment inside this `if` about when this happens

pkg/rollout/rollout.go
```diff
@@ -200,6 +208,41 @@ func (r *Rollout) PrepareRollback(svc *run.Service, stable, candidate string) *r
...
+		return nil, nil
+	case health.Healthy:
+		r.log.Debug("healthy candidate")
+		enoughTime, err := r.shouldRollForward(svc.Metadata.Annotations[LastRollForwardAnnotation], r.strategy.TimeBetweenRollForwardsMins)
+		if err != nil {
+			return nil, errors.Wrap(err, "could not determine if roll forward is allowed")
+		}
+		if !enoughTime {
+			r.log.Debug("no enough time since last roll forward, service unchanged")
+			return nil, nil
```
ahmetb, line 231: this should be inconclusive but we need to figure out how to return a human-readable explanation for the status annotation

pkg/rollout/rollout.go
```diff
@@ -290,13 +333,16 @@ func (r *Rollout) updateAnnotations(svc *run.Service, stable, candidate string)
...
+		svc.Metadata.Annotations[LastRollForwardAnnotation] = r.time.Now().Format(time.RFC3339)
 		return svc
 	}
 
 	svc.Metadata.Annotations[StableRevisionAnnotation] = stable
 	svc.Metadata.Annotations[CandidateRevisionAnnotation] = candidate
+	svc.Metadata.Annotations[LastRollForwardAnnotation] = r.time.Now().Format(time.RFC3339)
 	if r.shouldRollback {
 		svc.Metadata.Annotations[LastFailedCandidateRevisionAnnotation] = candidate
+		delete(svc.Metadata.Annotations, LastRollForwardAnnotation)
```
ahmetb, line 345: I'd keep it (for the records)

pkg/rollout/rollout.go
```diff
@@ -290,13 +333,16 @@ func (r *Rollout) updateAnnotations(svc *run.Service, stable, candidate string)
 	if r.promoteToStable {
 		svc.Metadata.Annotations[StableRevisionAnnotation] = candidate
 		delete(svc.Metadata.Annotations, CandidateRevisionAnnotation)
+		svc.Metadata.Annotations[LastRollForwardAnnotation] = r.time.Now().Format(time.RFC3339)
 		return svc
 	}
 
 	svc.Metadata.Annotations[StableRevisionAnnotation] = stable
 	svc.Metadata.Annotations[CandidateRevisionAnnotation] = candidate
+	svc.Metadata.Annotations[LastRollForwardAnnotation] = r.time.Now().Format(time.RFC3339)
```
ahmetb, line 342: init once `now := r.time.Now().Format(time.RFC3339)`

pkg/rollout/rollout.go
```diff
@@ -318,6 +364,23 @@ func (r *Rollout) diagnoseCandidate(candidate string, healthCriteria []config.Me
 	return d, errors.Wrap(err, "failed to diagnose candidate's health")
 }
 
+// shouldRollForward determines if enough time has passed since last roll
+// forward.
+func (r *Rollout) shouldRollForward(lastRollForwardStr string, timeBetweenRollForwards int) (bool, error) {
+	lastRollForward, err := time.Parse(time.RFC3339, lastRollForwardStr)
```
ahmetb, line 370: if annotation is empty, return a clear error saying X annotation is missing

pkg/rollout/rollout.go
```diff
@@ -318,6 +364,23 @@ func (r *Rollout) diagnoseCandidate(candidate string, healthCriteria []config.Me
 	return d, errors.Wrap(err, "failed to diagnose candidate's health")
 }
 
+// shouldRollForward determines if enough time has passed since last roll
+// forward.
+func (r *Rollout) shouldRollForward(lastRollForwardStr string, timeBetweenRollForwards int) (bool, error) {
```
ahmetb, line 369: `shouldRollForward` → `isTimeElapsed`

pkg/rollout/rollout.go
```diff
@@ -318,6 +364,23 @@ func (r *Rollout) diagnoseCandidate(candidate string, healthCriteria []config.Me
...
+	lastRollForward, err := time.Parse(time.RFC3339, lastRollForwardStr)
+	if err != nil {
+		return false, errors.Wrap(err, "failed to parse last roll forward time")
+	}
+
+	currentTime := r.time.Now()
+	r.log.Debugf("last roll forward: %v", lastRollForward)
+	r.log.Debugf("current time: %v", currentTime)
```
ahmetb, line 377: these are fields

ahmetb: lets move the log to the call site

pkg/rollout/rollout_test.go
```diff
@@ -148,13 +159,43 @@ func TestUpdateService(t *testing.T) {
...
+		{
+			name: "no enough time, do not roll forward",
+			traffic: []*run.TrafficTarget{
+				{RevisionName: "test-001", Percent: 100 - strategy.Steps[1], Tag: rollout.StableTag},
+				{RevisionName: "test-002", Percent: strategy.Steps[1], Tag: rollout.CandidateTag},
+				{LatestRevision: true, Tag: rollout.LatestTag},
+			},
+			annotations: map[string]string{
+				rollout.LastRollForwardAnnotation: clockMock.Now().Format(time.RFC3339),
+			},
```
ahmetb, line 181: create inline function


```
mkAnnotations := func(t time.Time, offset Duration) map[string]string { 
} 
```

inside the test function

pkg/rollout/rollout.go
gvso: Done!

pkg/rollout/rollout.go
gvso: Done!

pkg/rollout/rollout.go
gvso: Done!

pkg/rollout/rollout.go
gvso: Done!

pkg/rollout/rollout.go
gvso: I added a TODO for this

pkg/rollout/rollout.go
gvso: I renamed it to `hasEnoughTimeElapsed`. Is it fine?

pkg/rollout/rollout.go
gvso: Done!

pkg/rollout/rollout_test.go
gvso: Since it is also used in combination with other annotations, I created an utility function to generate the value instead.

pkg/rollout/rollout.go
ahmetb: I think the TODO above is a good idea.  Though I'm fine with it as it is right now.

pkg/rollout/rollout.go
```diff
@@ -285,16 +332,19 @@ func (r *Rollout) updateAnnotations(svc *run.Service, stable, candidate string)
...
 	if r.promoteToStable {
 		svc.Metadata.Annotations[StableRevisionAnnotation] = candidate
 		delete(svc.Metadata.Annotations, CandidateRevisionAnnotation)
+		svc.Metadata.Annotations[LastRolloutAnnotation] = now
 		return svc
 	}
 
 	svc.Metadata.Annotations[StableRevisionAnnotation] = stable
 	svc.Metadata.Annotations[CandidateRevisionAnnotation] = candidate
+	svc.Metadata.Annotations[LastRolloutAnnotation] = now
```
ahmetb, line 347: it seems like you always do this (in both cases). so why not just set it outside the `if` (e.g. right after `now:=...` ?

gvso: Done!

grayside:
Don't have time today, and this looks good at a glance :)

PR #64, opened 2020-08-07
stackdriver: Return 0 if no request in interval

gvso:
When no request was made in the interval, Latency and ErrorRate returns errors that are propagated all the way to the main package. No requests shouldn't represent a metrics error.

### DIFF
--- a/internal/stackdriver/stackdriver.go
+++ b/internal/stackdriver/stackdriver.go
@@ -96,6 +96,7 @@ func (p *Provider) RequestCount(ctx context.Context, offset time.Duration) (int6
 }
 
 // Latency returns the latency for the resource for the given offset.
+// It returns 0 if no request was made during the interval.
 func (p *Provider) Latency(ctx context.Context, offset time.Duration, alignReduceType metrics.AlignReduce) (float64, error) {
 	query := p.query.addFilter("metric.type", requestLatencies)
 	endTime := time.Now()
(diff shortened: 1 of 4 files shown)


gvso:
Here's my thought on this and why they should return 0:

1. We can make the argument that if you don't get any requests, your error rate is 0 since you didn't get any requests in the first place. A similar argument is applicable to latency.
2. If you got no requests in the interval, and you have some `min-request-count`, you will get an inconclusive result anyways.
3. If you specify `min-request-count=0`, we can probably assume the user wants a kitchen-timer mode. That is, the metrics don't matter for this user.
4 This behavior returns the same value as `RequestCount`, though there's an stronger and obvious argument for `RequestCount` returning 0 when no requests in interval.
5. I believe this approach is much easier to implement and think about

internal/stackdriver/stackdriver.go
```diff
@@ -188,27 +195,25 @@ func makeRequestForTimeSeries(logger *logrus.Entry, req *monitoring.ProjectsTime
 
 // calculateErrorResponseRate calculates the percentage of 5xx error response.
 //
-// It obtains all the successful responses (2xx) and the error responses (5xx),
-// add them up to form a 'total'. Then, it divides the number of error responses
-// by the total.
+// It gets all the server responses and calculates the error rate by performing
+// the operation (5xx responses / all responses). Then, it divides the number of
+// error responses by the total.
```
ahmetb, line 200: document behavior when 0 requests.

PR #65, opened 2020-08-10
rollout: Update traffic in UpdateService

gvso:
This creates a new file traffic.go and move all the traffic-related methods to the new file. It also refactors the methods to return traffic configuration, so the main caller, UpdateService, can update the service's traffic.

### DIFF
--- a/internal/rollout/rollout.go
+++ b/internal/rollout/rollout.go
@@ -132,7 +132,7 @@ func (r *Rollout) UpdateService(svc *run.Service) (*run.Service, error) {
 	// A new candidate does not have metrics yet, so it can't be diagnosed.
 	if isNewCandidate(svc, candidate) {
 		r.log.Debug("new candidate, assign some traffic")
-		svc = r.PrepareRollForward(svc, stable, candidate)
+		svc.Spec.Traffic = r.RollForwardTraffic(svc.Spec.Traffic, stable, candidate)
 		svc = r.updateAnnotations(svc, stable, candidate)
 		r.setHealthReportAnnotation(svc, "new candidate, no health report available yet")
 
@@ -146,17 +146,18 @@ func (r *Rollout) UpdateService(svc *run.Service) (*run.Service, error) {
 		return nil, errors.Wrapf(err, "failed to diagnose health for candidate %q", candidate)
 	}
 
-	svc, err = r.updateServiceBasedOnDiagnosis(svc, diagnosis.OverallResult, stable, candidate)
+	traffic, err := r.trafficBasedOnDiagnosis(svc, diagnosis.OverallResult, stable, candidate)
 	if err != nil {
-		return nil, errors.Wrap(err, "failed to update service after diagnosis")
+		return nil, errors.Wrap(err, "failed to configure traffic after diagnosis")
 	}
-	if svc == nil {
-		// If service was unchanged, nil is returned.
-		// TODO(gvso): This should go away once we start getting traffic config
-		// object from updateServiceBasedOnDiagnosis.
+	if traffic == nil {
+		// If traffic was unchanged, nil is returned.
+		// TODO(gvso): Do not return a nil service. Also update the annotations
+		// even if the traffic was unchanged.
 		return nil, nil
 	}
 
+	svc.Spec.Traffic = traffic
 	svc = r.updateAnnotations(svc, stable, candidate)
 	report := health.StringReport(r.strategy.HealthCriteria, diagnosis)
 	r.setHealthReportAnnotation(svc, report)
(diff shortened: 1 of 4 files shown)


pkg/rollout/rollout.go
```diff
@@ -148,17 +148,18 @@ func (r *Rollout) UpdateService(svc *run.Service) (*run.Service, error) {
...
 	}
 
-	svc, err = r.updateServiceBasedOnDiagnosis(svc, diagnosis.OverallResult, stable, candidate)
+	traffic, err := r.trafficBasedOnDiagnosis(svc, diagnosis.OverallResult, stable, candidate)
 	if err != nil {
-		return nil, errors.Wrap(err, "failed to update service after diagnosis")
+		return nil, errors.Wrap(err, "failed to configure traffic after diagnosis")
 	}
-	if svc == nil {
-		// If service was unchanged, nil is returned.
```
gvso, line 156: I left it as nil instead of returning the same, unchanged traffic. This is because we need to differentiate between an unchanged traffic because of inconclusive/no enough time since last rollout. So, it is easy to check for that if we get a `nil` traffic.

I think we could also use `go-cmp` to compare the old traffic to the new one to determine if the traffic was unchanged.

pkg/rollout/rollout.go
```diff
@@ -148,17 +148,18 @@ func (r *Rollout) UpdateService(svc *run.Service) (*run.Service, error) {
...
-		// TODO(gvso): This should go away once we start getting traffic config
-		// object from updateServiceBasedOnDiagnosis.
```
gvso, line 158: I left it as nil instead of returning the same, unchanged traffic. This is because we need to differentiate between an unchanged traffic because of inconclusive/no enough time since last rollout. So, it is easy to check for that if we get a nil traffic.

We could also use go-cmp to compare the old traffic to the new one to determine if the traffic was unchanged if using nil doesn't sound like a good idea.

pkg/rollout/traffic.go
```diff
@@ -0,0 +1,184 @@
...
+	"github.com/GoogleCloudPlatform/cloud-run-release-operator/pkg/health"
+	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
+	"google.golang.org/api/run/v1"
+)
+
+// trafficBasedOnDiagnosis updates a service's traffic configuration
+// based on the diagnosis.
+// If traffic should not changed, nil is returned.
+func (r *Rollout) trafficBasedOnDiagnosis(svc *run.Service, diagnosis health.DiagnosisResult, stable, candidate string) ([]*run.TrafficTarget, error) {
```
ahmetb, line 15: determineTraffic["",Targets,Rules] any of these are better names than this one IMO

pkg/rollout/traffic.go
```diff
@@ -0,0 +1,184 @@
...
+	switch diagnosis {
+	case health.Inconclusive:
+		r.log.Debug("health check inconclusive")
+		return nil, nil
+	case health.Healthy:
+		r.log.Debug("healthy candidate")
+		lastRollout := svc.Metadata.Annotations[LastRolloutAnnotation]
+		enoughTime, err := r.hasEnoughTimeElapsed(lastRollout, r.strategy.TimeBetweenRollouts)
+		if err != nil {
+			return nil, errors.Wrap(err, "could not determine if roll out is allowed")
```
ahmetb, line 25: be more descriptive: `error while determining enough time elapsed`

pkg/rollout/traffic.go
```diff
@@ -0,0 +1,184 @@
...
+		return nil, errors.Errorf("invalid candidate's health diagnosis %v", diagnosis)
+	}
+}
+
+// RollForwardTraffic updates the traffic configuration to increase the traffic
+// to the candidate.
+//
+// It creates new traffic configurations for the candidate and stable revisions
+// and respects user-defined revision tags.
+func (r *Rollout) RollForwardTraffic(traffic []*run.TrafficTarget, stable, candidate string) []*run.TrafficTarget {
```
ahmetb, line 47: these probably should be unexported since the only place they're called from are unexported.

pkg/rollout/rollout.go
ahmetb: go-cmp is a bad idea.
However returning `(Traffic, bool)` to indicate a change was made or not is more go-idiomatic.

ahmetb: That said you can do that in a separate PR.

pkg/rollout/traffic.go
```diff
@@ -0,0 +1,183 @@
...
+	}
+}
+
+// rollForwardTraffic updates the traffic configuration to increase the traffic
+// to the candidate.
+//
+// It creates new traffic configurations for the candidate and stable revisions
+// and respects user-defined revision tags.
+func (r *Rollout) rollForwardTraffic(traffic []*run.TrafficTarget, stable, candidate string) []*run.TrafficTarget {
+	r.log.Debug("splitting traffic")
```
ahmetb, line 47: seems redundant here right now

pkg/rollout/rollout.go
gvso: Should this function return 3 values when that's implemented `(Traffic, bool, error)`?

ahmetb: I wouldn't mind, but if traffic==nil implies bool=false, then it's an overkill to have bool.

gvso: If we add a boolean, we can just return the old traffic with no changes, so we will never have to check `if traffic = nil` since we guarantee a traffic object regardless of the diagnosis/situation.

PR #66, opened 2020-08-10
main: Use time.Duration for time-related flags

gvso:
This updates the type expected by the `cli-run-interval` and `healthcheck-offset` flags to time.Duration. This makes it more logical to the user and simplify our current implementation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -190,9 +190,8 @@ The rollout strategy consists of the steps and health criteria.
 
 - `-cli-run-interval`: The time between each health check (default: `60s`). This
 is only need it if running with `-cli` option.
-- `-healthcheck-offset`: The range of time for retrieving metrics relative to
-the current rollout process (default: `30m`). Default value means that metrics
-from the last 30 minutes are retrieved.
+- `-healthcheck-offset`: Time window to look back during health check to assess
+the candidate revision's health (default: `30m`).
 - `-min-requests`: The minimum number of requests needed to determine the
 candidate's health (default: `100`)
 - `-min-wait`: The minimum time before rolling out further (default: `30m`)
--- a/README.md
+++ b/README.md
@@ -188,10 +188,10 @@ available Cloud Run regions](https://cloud.google.com/run/docs/locations))
 
 The rollout strategy consists of the steps and health criteria.
 
-- `-cli-run-interval`: The time between each health check, in seconds (default:
-`60`). This is only need it if running with `-cli` option.
-- `-healthcheck-offset`: To evaluate the candidate's health, use metrics from
-the last `N` minutes relative to current rollout process (default: `30`)
+- `-cli-run-interval`: The time between each health check (default: `60s`). This
+is only need it if running with `-cli` option.
+- `-healthcheck-offset`:range of time for retrieving metrics relative to current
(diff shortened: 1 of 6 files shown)


ahmetb:
Still need to validate negative times which are negative inputs to our program :) 
https://play.golang.org/p/fkj1tmcd8fQ

gvso:
`cli-run-interval` was never checked before, which is actually a bug. So, I agree we should check it now.

The health check offset is checked in `config.Validate`. Should we check it before?

ahmetb:
Overall, would be definitely good to check these as early as possible (e.g. before a server starts etc)

gvso:
The configuration is always validated right after the flags are validated. I added validation for both flags though.

README.md
```diff
@@ -91,10 +91,10 @@ available Cloud Run regions](https://cloud.google.com/run/docs/locations))
...
 
-- `-cli-run-interval`: The time between each health check, in seconds (default:
-`60`). This is only need it if running with `-cli` option.
-- `-healthcheck-offset`: To evaluate the candidate's health, use metrics from
-the last `N` minutes relative to current rollout process (default: `30`)
+- `-cli-run-interval`: The time between each health check (default: `60s`). This
+is only need it if running with `-cli` option.
+- `-healthcheck-offset`:range of time for retrieving metrics relative to current
+rollout process (default: `30m`)
 - `-min-requests`: The minimum number of requests needed to determine the
```
ahmetb, line 99: probably make a note that 

> [...] needed (with -healthcheck-offset) to determine

also worth adding to cli help

README.md
```diff
@@ -91,10 +91,10 @@ available Cloud Run regions](https://cloud.google.com/run/docs/locations))
 
 The rollout strategy consists of the steps and health criteria.
 
-- `-cli-run-interval`: The time between each health check, in seconds (default:
-`60`). This is only need it if running with `-cli` option.
-- `-healthcheck-offset`: To evaluate the candidate's health, use metrics from
-the last `N` minutes relative to current rollout process (default: `30`)
+- `-cli-run-interval`: The time between each health check (default: `60s`). This
+is only need it if running with `-cli` option.
+- `-healthcheck-offset`:range of time for retrieving metrics relative to current
```
ahmetb: I'd probably rewrite this as 

> Duration to look back for metrics to assess health of the candidate revision.

or

> time window to look back at the time of health check to assess health of the candidate revision.

ahmetb:
FYI  56a973f commit description doesn't match the diff

README.md
gvso: I like second more but would rephrase it to

> time window to look back during health check to assess the candidate revision's health.

to avoid the repetition of the word "time" in different contexts and to separate the occurrences of "health" a little bit more.

README.md
gvso: What exactly do you mean by this? What's needed with `healthcheck-offset`?

ahmetb: I meant that it is worth noting that min-requests are calculated "within $healthcheck-offset minutes" and not all-time.

gvso: Oh sure. Let me add that piece

PR #67, opened 2020-08-10
config: Allow multiple strategies

gvso:
As of now, the same rollout strategy was applied to all the targeted services. This makes changes to the configuration to allow different strategies for different targets.


gvso:
The configuration for the application looks as follow (YAML-formatted):

```yaml
strategies:
- target:
    project: <YOUR_PROJECT>
    regions: [region1, region2]
    labelSelector: "team=backend"
  minWait: 5m
  healthCheckOffset: 30m
  health:
  - metric: min-requests
    threshold: 300
  - metric: latency_p99
    threshold: 1000
  - metric: error-rate
    threshold: 0.01
```

PR #72, opened 2020-08-12
main: Autodetect project if one is not specified

gvso:
This uses the compute metadata package to detect the ID of the project on which the service is deployed. If the tool is not being run on GCE, it tries to detect the project ID by executing a `gcloud` command to get the default project

### DIFF
--- a/cmd/operator/main.go
+++ b/cmd/operator/main.go
@@ -15,15 +15,18 @@
 package main
 
 import (
+	"bytes"
 	"context"
 	"flag"
 	"fmt"
 	"net/http"
 	"os"
+	"os/exec"
 	"strconv"
 	"strings"
 	"time"
 
+	"cloud.google.com/go/compute/metadata"
 	"github.com/GoogleCloudPlatform/cloud-run-release-operator/internal/config"
 	"github.com/GoogleCloudPlatform/cloud-run-release-operator/internal/metrics"
 	"github.com/GoogleCloudPlatform/cloud-run-release-operator/internal/metrics/sheets"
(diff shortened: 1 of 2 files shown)


cmd/operator/main.go
```diff
@@ -237,3 +250,28 @@ func printHealthCriteria(logger *logrus.Logger, healthCriteria []config.HealthCr
...
+	if metadata.OnGCE() {
+		v, err := metadata.ProjectID()
+		if err != nil {
+			return "", errors.Wrapf(err, "error when getting project ID from compute metadata: %v", err)
+		}
+		return v, nil
+	}
+
+	// Try to get project ID by retrieving default value in gcloud.
+	cmd := exec.Command("gcloud", "config", "get-value", "core/project")
```
ahmetb, line 264: add `-q` to make it not show potential prompts

cmd/operator/main.go
```diff
@@ -237,3 +250,28 @@ func printHealthCriteria(logger *logrus.Logger, healthCriteria []config.HealthCr
...
+	var out bytes.Buffer
+	cmd.Stdout = &out
+	err := cmd.Run()
+	if err != nil {
+		return "", errors.Wrap(err, "error when running gcloud command to get default project")
```
ahmetb, line 269: would be good to return the error message as well. so capture stderr in another buffer

cmd/operator/main.go
```diff
@@ -237,3 +250,28 @@ func printHealthCriteria(logger *logrus.Logger, healthCriteria []config.HealthCr
...
+	}
+
+	v := strings.Trim(out.String(), "\n")
```
ahmetb, line 272: strings.TrimSpace

cmd/operator/main.go
```diff
@@ -251,25 +251,31 @@ func printHealthCriteria(logger *logrus.Logger, healthCriteria []config.HealthCr
...
 
 	// Try to get project ID by retrieving default value in gcloud.
-	cmd := exec.Command("gcloud", "config", "get-value", "core/project")
-	var out bytes.Buffer
-	cmd.Stdout = &out
+	cmd := exec.Command("gcloud", "config", "get-value", "core/project", "-q")
+	var stdout bytes.Buffer
+	var stderr bytes.Buffer
+	cmd.Stdout = &stdout
+	cmd.Stderr = &stderr
```
ahmetb, line 268: not necessary but you can do this in the future:

var stdout, stderr bytes.Buffer
cmd.Stdout, cmd.Stderr = &stdout, &stderr

gvso: I made this change too

PR #79, opened 2020-08-12
rollout: Indicate if traffic config was changed

gvso:
This adds a boolean return value to explicitly indicate if the traffic configuration was changed during the update of the service. Indicating that the traffic changed would be useful when we add annotations to all services that have stable and candidate revisions. That is, all calls to UpdateService are going to update at least the annotation part given that there are stable and candidate revisions.

### DIFF
--- a/internal/rollout/rollout.go
+++ b/internal/rollout/rollout.go
@@ -104,28 +104,33 @@ func (r *Rollout) Rollout() (bool, error) {
 		"region":  r.region,
 	})
 
-	svc, err := r.UpdateService(r.service)
+	_, trafficChanged, err := r.UpdateService(r.service)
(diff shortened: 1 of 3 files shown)


internal/rollout/rollout.go
```diff
@@ -104,28 +104,33 @@ func (r *Rollout) Rollout() (bool, error) {
 		"region":  r.region,
 	})
 
-	svc, err := r.UpdateService(r.service)
+	_, trafficChanged, err := r.UpdateService(r.service)
```
gvso, line 107: I'm not using the returned service here, but it's useful for testing since we can test that we got the expected changes, not only that the traffic has changed.

PR #80, opened 2020-08-12
Image build instructions

ahmetb:
We're probably not gonna publish off-the-shelf images at this
point, so let's let users build from the source.

### DIFF
--- a/README.md
+++ b/README.md
@@ -132,12 +132,16 @@ To set up this on Cloud Run, run the following steps on your shell:
          --role=roles/monitoring.viewer
     ```
 
-1. (Optional) Mirror the docker image to your GCP project.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -131,7 +135,7 @@ To set up this on Cloud Run, run the following steps on your shell:
         --region=us-central1 \
         --image=gcr.io/$PROJECT_ID/cloud-run-release-manager \
         --service-account=release-manager@${PROJECT_ID}.iam.gserviceaccount.com \
-        --args=-project=$PROJECT_ID
+        --args=-project=$PROJECT_ID --args=-verbosity=debug
```
gvso, line 138: We can probably remove the project flag now, but we can do it later

PR #87, opened 2020-08-13
main: Log more info when autodetecting project ID

gvso:


cmd/operator/main.go
```diff
@@ -109,6 +109,11 @@ func init() {
 	flag.StringVar(&flGoogleSheetsID, "google-sheets", "", "ID of public Google sheets document to use as metrics provider")
 	flag.Parse()
 
+	args := flag.Args()
```
ahmetb, line 112: it seems like this is leftover

PR #88, opened 2020-08-13
main: Do not silently accept positional arguments

gvso:


cmd/operator/main.go
```diff
@@ -109,6 +109,11 @@ func init() {
 	flag.StringVar(&flGoogleSheetsID, "google-sheets", "", "ID of public Google sheets document to use as metrics provider")
 	flag.Parse()
 
+	args := flag.Args()
+	if len(args) != 0 {
+		logrus.Fatalf("invalid command args %v", args)
```
ahmetb, line 114: that's not a great message. just say:

> positional arguments not accepted: %#v

gvso: Done!

PR #90, opened 2020-08-13
Remove paddings in diagrams about rollout/rollback

gvso:


gvso:
See how the [README](https://github.com/gvso/cloud-run-release-operator/blob/remove-diagram-padding/README.md) looks

PR #94, opened 2020-08-14
main: Print flags

gvso:
This is with the purpose of showing more detail to the user about the configuration values for debug purposes. For now, we can print the command line flags

### DIFF
--- a/cmd/operator/main.go
+++ b/cmd/operator/main.go
@@ -142,6 +142,7 @@ func main() {
 	if err := validateFlags(); err != nil {
 		logger.Fatalf("invalid flags: %v", err)
 	}
+	logger.Debug(flagsToString())
 
 	// Configuration.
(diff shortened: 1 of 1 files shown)


cmd/operator/main.go
```diff
@@ -145,6 +145,7 @@ func main() {
 	if err := validateFlags(); err != nil {
 		logger.Fatalf("invalid flags: %v", err)
 	}
+	logger.Debug(flagsToString())
```
ahmetb, line 148: these also could be better as fields and msg="parsed configuration"

gvso: That's true. The only thing about making them fields is that they will be alphabetically sorted instead of having related values together

ahmetb: Yeah lets go with the current version for now

Issue #75, opened 2020-08-12
Publish to Pub/Sub after a rollout/rollback have occurred


gvso:
In order to allow other functionality around the Release Manager, we need to provide a way to subscribe to events. This will allow things like smoke tests and notification features. The following format could be useful:

```
{
	"event": "rollout", // or "rollback"
	"candidateRevisionName": "hello-002-abc",
	"candidateRevisionPercent": 50,
	"candidateRevisionURL": "https://candidate---hello-wh32u62c4l-ue.a.run.app",
        "candidateWasPromotedToStable": false,
	"service": {Service JSON}
}
```

ahmetb:
@steren wdyt of the event format above? We want to optionally do:

     --notify-pubsub=projects/x/topics/y

and let users push rollout and rollback notifications for them to be able to run smoke tests (or do other things like ChatOps). Is the information above sufficient for those cases?

steren:
Seems like a good idea. Trying to build something real with these is the best way to see if the event contains the info you would need.

gvso:
@ahmetb Why do we need to pass the project as part of the flag? We could use the project ID we already autodetect (or asked for `-project`)?

ahmetb:
It's possible to push to topics in another project and it's a common use case.

PR #95, opened 2020-08-14
pubsub: Add pubsub package for publishing events

gvso:
This adds a pubsub package that adds a wrapper around the Google Cloud Pub/Sub package to publish rollout events

### DIFF
--- a/internal/notification/pubsub/pubsub.go
+++ b/internal/notification/pubsub/pubsub.go
@@ -7,26 +7,31 @@ import (
 	"regexp"
 
 	cloudpubsub "cloud.google.com/go/pubsub"
-	"github.com/GoogleCloudPlatform/cloud-run-release-operator/internal/health"
-	"github.com/GoogleCloudPlatform/cloud-run-release-operator/internal/util"
+	"github.com/GoogleCloudPlatform/cloud-run-release-manager/internal/health"
+	"github.com/GoogleCloudPlatform/cloud-run-release-manager/internal/util"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 	"google.golang.org/api/run/v1"
 )
 
+// Event types.
+const (
+	rolloutEvent  = "rollout"
+	rollbackEvent = "rollback"
+)
+
 // Client represents a client to Google Cloud Pub/Sub.
 type Client interface {
-	Publish(ctx context.Context, msg Message) error
+	Publish(ctx context.Context, event RolloutEvent) error
 }
 
 // PubSub is a Google Cloud Pub/Sub client to publish messages.
 type PubSub struct {
-	client *cloudpubsub.Client
-	topic  *cloudpubsub.Topic
+	topic *cloudpubsub.Topic
 }
 
-// Message is the format when publishing to Pub/Sub.
-type Message struct {
+// RolloutEvent is the format of an event published to Pub/Sub.
+type RolloutEvent struct {
 	Event                        string       `json:"event"`
 	CandidateRevisionName        string       `json:"candidateRevisionName"`
 	CandidateRevisionPercent     int          `json:"candidateRevisionPercent"`
@@ -35,7 +40,7 @@ type Message struct {
 	Service                      *run.Service `json:"service"`
 }
 
-// New initializes a PubSub client.
+// New initializes a PubSub client to a topic in a project.
 func New(ctx context.Context, projectID string, topicName string) (ps PubSub, err error) {
 	logger := util.LoggerFrom(ctx)
 	client, err := cloudpubsub.NewClient(ctx, projectID)
@@ -49,19 +54,20 @@ func New(ctx context.Context, projectID string, topicName string) (ps PubSub, er
 	}
 	project := match[1]
 	topicID := match[2]
-	logger.WithFields(logrus.Fields{"topicProject": project, "topicID": topicID}).Debug("parsed topic name value")
+	logger.WithFields(logrus.Fields{"topicProject": project, "topicID": topicID}).Debug("parsed pubsub topic configuration")
 
 	return PubSub{
-		client: client,
-		topic:  client.TopicInProject(topicID, project),
+		topic: client.TopicInProject(topicID, project),
 	}, nil
 }
 
-// NewMessage initializes a message.
-func NewMessage(svc *run.Service, diagnosis health.DiagnosisResult, candidateWasPromoted bool) (Message, error) {
-	event := "rollout"
+// NewRolloutEvent initializes an event to publish to PubSub.
+//
+// svc must be the updated Service instance as the result of the rollout.
+func NewRolloutEvent(svc *run.Service, diagnosis health.DiagnosisResult, candidateWasPromoted bool) (RolloutEvent, error) {
+	event := rolloutEvent
 	if diagnosis == health.Unhealthy {
-		event = "rollback"
+		event = rollbackEvent
 	}
 
 	var candidateRevision *run.TrafficTarget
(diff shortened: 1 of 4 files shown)


gvso:
This is related to #75

internal/pubsub/pubsub.go
```diff
@@ -0,0 +1,138 @@
...
+}
+
+// PubSub is a Google Cloud Pub/Sub client to publish messages.
+type PubSub struct {
+	client *cloudpubsub.Client
+	topic  *cloudpubsub.Topic
+}
+
+// Message is the format when publishing to Pub/Sub.
+type Message struct {
```
ahmetb, line 29: hmm the best way probably would be to have pkg/notification, and have this be called `RolloutEvent` and pkg/notification/pubsub would be a sub-type of that. 

but for now it's ok

internal/pubsub/pubsub.go
```diff
@@ -0,0 +1,138 @@
...
+)
+
+// Client represents a client to Google Cloud Pub/Sub.
+type Client interface {
+	Publish(ctx context.Context, msg Message) error
+}
+
+// PubSub is a Google Cloud Pub/Sub client to publish messages.
+type PubSub struct {
+	client *cloudpubsub.Client
```
ahmetb, line 24: you don't use client anywhere in the code.

internal/pubsub/pubsub.go
```diff
@@ -0,0 +1,138 @@
...
+	}
+
+	var candidateRevision *run.TrafficTarget
+	var err error
+	if candidateWasPromoted {
+		// The last candidate is now the stable.
+		candidateRevision, err = findRevisionWithTag(svc, "stable")
+	} else {
+		candidateRevision, err = findRevisionWithTag(svc, "candidate")
+	}
```
ahmetb, line 74: it's a bit weird we re-evaluate the type and compute things again here.

ahmetb: don't we have these already computed in rollout code?

internal/pubsub/pubsub.go
```diff
@@ -0,0 +1,138 @@
...
+	data, err := json.Marshal(msg)
+	if err != nil {
+		return errors.Wrap(err, "failed to marshal message")
+	}
+
+	logger := util.LoggerFrom(ctx)
+	ps.topic.Publish(ctx, &cloudpubsub.Message{
+		Data: data,
+	})
+	logger.WithField("size", len(data)).Debug("published data to Pub/Sub")
```
ahmetb, line 100: let's use "event" instead of "data" or "message"

internal/pubsub/pubsub.go
gvso: We do, but we only get the revision name. There's also the problem that the info about the revision is obtained before rolling out/back, so it became outdated by the time things are published to PubSub.

That's without mentioning the problem outlined in the comment for `findRevisionWithTag`

internal/notification/pubsub/pubsub.go
```diff
@@ -0,0 +1,136 @@
...
+	logger.WithFields(logrus.Fields{"topicProject": project, "topicID": topicID}).Debug("parsed topic name values")
+
+	return PubSub{
+		topic: client.TopicInProject(topicID, project),
+	}, nil
+}
+
+// NewRolloutEvent initializes an event to publish to PubSub.
+func NewRolloutEvent(svc *run.Service, diagnosis health.DiagnosisResult, candidateWasPromoted bool) (RolloutEvent, error) {
+	event := "rollout"
```
ahmetb, line 60: make these strings as const ideally

internal/notification/pubsub/pubsub.go
```diff
@@ -0,0 +1,136 @@
...
+
+// Publish publishes message to the topic.
+func (ps PubSub) Publish(ctx context.Context, event RolloutEvent) error {
+	data, err := json.Marshal(event)
+	if err != nil {
+		return errors.Wrap(err, "failed to marshal message")
+	}
+
+	logger := util.LoggerFrom(ctx)
+	ps.topic.Publish(ctx, &cloudpubsub.Message{
```
ahmetb, line 95: handle err

internal/notification/pubsub/pubsub.go
```diff
@@ -0,0 +1,136 @@
...
+	Event                        string       `json:"event"`
+	CandidateRevisionName        string       `json:"candidateRevisionName"`
+	CandidateRevisionPercent     int          `json:"candidateRevisionPercent"`
+	CandidateRevisionURL         string       `json:"candidateRevisionURL"`
+	CandidateWasPromotedToStable bool         `json:"candidateWasPromotedToStable"`
+	Service                      *run.Service `json:"service"`
+}
+
+// New initializes a PubSub client to a topic in a project.
+func New(ctx context.Context, projectID string, topicName string) (ps PubSub, err error) {
```
ahmetb, line 38: this should return a `func()` to "stop or wait flushing"

internal/notification/pubsub/pubsub.go
```diff
@@ -0,0 +1,136 @@
...
+		return ps, errors.Wrap(err, "failed to initialize Pub/Sub client")
+	}
+
+	match := regexp.MustCompile(`projects/([^/]*)/topics/([^/]*)`).FindStringSubmatch(topicName)
+	if len(match) != 3 {
+		return ps, errors.Errorf("invalid topic name %s", topicName)
+	}
+	project := match[1]
+	topicID := match[2]
+	logger.WithFields(logrus.Fields{"topicProject": project, "topicID": topicID}).Debug("parsed topic name values")
```
ahmetb, line 51: `parsed pubsub topic configuration`

internal/notification/pubsub/pubsub.go
```diff
@@ -0,0 +1,136 @@
...
+
+	return PubSub{
+		topic: client.TopicInProject(topicID, project),
+	}, nil
+}
+
+// NewRolloutEvent initializes an event to publish to PubSub.
```
ahmetb, line 58: add: "svc must be the updated Service instance as the result of the rollout"

internal/notification/pubsub/pubsub.go
```diff
@@ -0,0 +1,136 @@
...
+	}
+	if target == nil {
+		return nil, errors.Errorf("could not find revision with tag %s", tag)
+	}
+
+	url, err := url.Parse(svc.Status.Url)
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to parse the service's url %s", svc.Status.Url)
+	}
+	url.Host = tag + "---" + url.Host
```
ahmetb, line 132: can't do this, it's Cloud Run (fully managed) specific.

ahmetb: add a // TODO  saying this won't work for Cloud Run for Anthos.

internal/notification/pubsub/pubsub.go
gvso: Should we return a third value or can we just add a wrapper for `Stop` in our package?

internal/notification/pubsub/pubsub.go
gvso: As discussed, no errors are returned in [`Publish`](https://pkg.go.dev/cloud.google.com/go/pubsub?tab=doc#Topic.Publish) unless we call [`Get`](https://pkg.go.dev/cloud.google.com/go/pubsub?tab=doc#PublishResult.Get) on the result (blocking operation). If we're expecting `Stop` to be called at the end, calling `Get` seems unnecessary.

internal/notification/pubsub/pubsub.go
```diff
@@ -97,10 +103,19 @@ func (ps PubSub) Publish(ctx context.Context, msg Message) error {
...
 }
 
+// Stop is a wrapper around Cloud Run Pub/Sub package's Stop method on Topic.
+//
+// It sends all remaining published messages and stop goroutines created for
+// handling publishing. Returns once all outstanding messages have been sent or
+// have failed to be sent.
+func (ps PubSub) Stop() {
+	ps.topic.Stop()
+}
```
ahmetb, line 117: I was imagining this would be returned as a `func()` from `New()`. It would be harder to ignore or miss that way.

gvso: That's also possible (and asked about in my previous comment). Let's see how this looks in the next PR as we implement the package in the rollout

PR #101, opened 2020-08-17
rollout: Change message when candidate is not found

gvso:
This updates the log messages when no stable or candidate revision is found. It also changes the type of log to debug in the case of no candidate since it occurs often

### DIFF
--- a/internal/rollout/rollout.go
+++ b/internal/rollout/rollout.go
@@ -118,13 +118,13 @@ func (r *Rollout) Rollout() (bool, error) {
 func (r *Rollout) UpdateService(svc *run.Service) (*run.Service, error) {
 	stable := DetectStableRevisionName(svc)
 	if stable == "" {
-		r.log.Info("could not determine stable revision")
+		r.log.Info("no stable revision found")
 		return nil, nil
 	}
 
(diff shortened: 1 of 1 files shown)


internal/rollout/rollout.go
```diff
@@ -118,13 +118,13 @@ func (r *Rollout) Rollout() (bool, error) {
 func (r *Rollout) UpdateService(svc *run.Service) (*run.Service, error) {
 	stable := DetectStableRevisionName(svc)
 	if stable == "" {
-		r.log.Info("could not determine stable revision")
+		r.log.Info("no stable revision found")
```
ahmetb, line 121: cannot find a stable revision (that gets 100% of the traffic)

internal/rollout/rollout.go
```diff
@@ -118,13 +118,13 @@ func (r *Rollout) Rollout() (bool, error) {
...
 		return nil, nil
 	}
 
 	candidate := DetectCandidateRevisionName(svc, stable)
 	if candidate == "" {
-		r.log.Info("could not determine candidate revision")
+		r.log.Debug("no candidate revision found")
```
ahmetb, line 127: currently no candidate revisions exist to roll forward

PR #103, opened 2020-08-18
main: Default min requests to 0

gvso:
The previous value of 100 is an issue since it requires that the service gets a lot of requests, especially at the beginning of the rollout, to determine the candidate's health

### DIFF
--- a/README.md
+++ b/README.md
@@ -127,18 +127,12 @@ When running the Release Manager, default configuration values are used. The
 most relevant values are:
 
 - `-healthcheck-offset=30m`: When assessing the candidate's health, the Release
-  Manager will get metrics about the candidate revision from the last 30
-  minutes
-- `-min-requests=100`: The Release Manager expects `100` requests in the past 30
-  minutes (given by `healthcheck-offset`). If no enough requests were made in
-  that interval, no rollout or rollback is performed even if the other metrics
-  show a healthy/unhealthy candidate
+  Manager will get metrics about the candidate revision from the last 30 minutes
 - `-max-error-rate=1`: By default, the only health criteria is a expected max
   server error rate of 1%
-- `-min-wait=30m`: If the candidate revision got the minimum number of requests
-  and was healthy, it can be rolled out further only if 30 minutes since last
-  roll out has elapsed. If the candidate was unhealthy, however, it is rolled
-  back independent of the elapsed time since last rollout.
+- `-min-wait=30m`: If the candidate revision is healthy, it can be rolled out
+  further only if 30 minutes since last roll out has elapsed. If the candidate
+  was unhealthy, however, it is rolled back independent of the elapsed time.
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -127,18 +127,12 @@ When running the Release Manager, default configuration values are used. The
 most relevant values are:
 
 - `-healthcheck-offset=30m`: When assessing the candidate's health, the Release
-  Manager will get metrics about the candidate revision from the last 30
-  minutes
-- `-min-requests=100`: The Release Manager expects `100` requests in the past 30
-  minutes (given by `healthcheck-offset`). If no enough requests were made in
-  that interval, no rollout or rollback is performed even if the other metrics
-  show a healthy/unhealthy candidate
```
ahmetb, line 135: I think we should still specify the default as 0

ahmetb:
I think the goal was to add the defaults explicitly to deployment command?

gvso:
Yes. I went ahead and just pass the same values as the default ones, so the experience in CLI is the same as the deployed one by default. Any thoughts on that?

README.md
```diff
@@ -210,7 +210,7 @@ shell:
         --region=us-central1 \
         --image=gcr.io/$PROJECT_ID/cloud-run-release-manager \
         --service-account=release-manager@${PROJECT_ID}.iam.gserviceaccount.com \
-        --args=-verbosity=debug
+        --args=-verbosity=debug,-healthcheck-offset=30m,-min-requests=0,-max-error-rate=1,-min-wait=30m
```
ahmetb, line 213: split these into separate --args=

README.md
```diff
@@ -127,18 +127,18 @@ When running the Release Manager, default configuration values are used. The
...
-  that interval, no rollout or rollback is performed even if the other metrics
-  show a healthy/unhealthy candidate
+  Manager will get metrics about the candidate revision from the last 30 minutes
+- `-min-requests=0`: The Release Manager expects this number of requests in the
+  past 30 minutes (given by `healthcheck-offset`). If no enough requests were
+  made in that interval, no rollout or rollback is performed even if the other
+  metrics show a healthy/unhealthy candidate. The default value of `0` means
+  that the number of requests should be ignored when the Release Manager makes
+  a decission about rolling out/back.
 - `-max-error-rate=1`: By default, the only health criteria is a expected max
```
ahmetb, line 137: would be also good to add `(in percent)` here

README.md
```diff
@@ -127,18 +127,18 @@ When running the Release Manager, default configuration values are used. The
 most relevant values are:
 
 - `-healthcheck-offset=30m`: When assessing the candidate's health, the Release
```
ahmetb, line 129: it's almost like we should document these bullet points right after deploying with these params.

gvso: We could add a link to the section "Default rollout configuration"

PR #104, opened 2020-08-18
rollout: Report health for all cases

gvso:
This reports the health for all possible cases. In particular, it reports inconclusive diagnosis (no enough requests) and healthy diagnosis without enough elapsed time.

### DIFF
--- a/internal/health/health.go
+++ b/internal/health/health.go
@@ -82,16 +82,20 @@ func Diagnose(ctx context.Context, healthCriteria []config.HealthCriterion, actu
 		}
 
 		isMet := isCriteriaMet(criteria.Metric, criteria.Threshold, value)
+		result := CheckResult{Threshold: criteria.Threshold, ActualValue: value}
 
 		// For unmet request count, return inconclusive and empty results.
 		if !isMet && criteria.Metric == config.RequestCountMetricsCheck {
 			logger.Debug("unmet criterion")
 			diagnosis = Inconclusive
-			results = nil
+			// TODO: This will be an issue if the request count is not the first
+			// criterion in the array since the caller can assume the order
+			// of the check results is the same as the order of the health
+			// criteria.
+			results = []CheckResult{result}
 			break
 		}
 
-		result := CheckResult{Threshold: criteria.Threshold, ActualValue: value}
 		if !isMet {
 			logger.Debug("unmet criterion")
 			diagnosis = Unhealthy
--- a/internal/health/health.go
+++ b/internal/health/health.go
@@ -54,11 +54,15 @@ type CheckResult struct {
 // actual values are not the same, the diagnosis is Unknown and an error is
 // returned.
 //
-// If the minimum number of requests is not met, then health cannot be
-// determined and diagnosis is Inconclusive.
+// Otherwise, all metrics criteria are checked to determine the diagnosis:
+// healthy, unhealthy, or inconclusive.
 //
-// Otherwise, all metrics criteria are checked to determine whether the revision
-// is healthy or not.
+// If the minimum number of requests is not met, the diagnosis is Inconclusive
+// even though all other criteria are met.
+//
+// However, if any criteria other than the request count is not met, the
+// diagnosis is unhealthy independent on the request count criteria. That is,
+// Unhealthy has precedence over Inconclusive.
 func Diagnose(ctx context.Context, healthCriteria []config.HealthCriterion, actualValues []float64) (Diagnosis, error) {
 	logger := util.LoggerFrom(ctx)
 	if len(healthCriteria) != len(actualValues) {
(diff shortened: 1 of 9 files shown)


internal/health/health.go
```diff
@@ -82,16 +82,20 @@ func Diagnose(ctx context.Context, healthCriteria []config.HealthCriterion, actu
...
 
 		isMet := isCriteriaMet(criteria.Metric, criteria.Threshold, value)
+		result := CheckResult{Threshold: criteria.Threshold, ActualValue: value}
 
 		// For unmet request count, return inconclusive and empty results.
 		if !isMet && criteria.Metric == config.RequestCountMetricsCheck {
 			logger.Debug("unmet criterion")
 			diagnosis = Inconclusive
-			results = nil
+			// TODO: This will be an issue if the request count is not the first
```
gvso, line 91: Potentially solvable easily if we store Metric as field in `CheckResult`

internal/health/health.go
```diff
@@ -82,16 +82,20 @@ func Diagnose(ctx context.Context, healthCriteria []config.HealthCriterion, actu
...
+			// criterion in the array since the caller can assume the order
+			// of the check results is the same as the order of the health
+			// criteria.
+			results = []CheckResult{result}
```
gvso, line 95: Potentially solvable easily if we store Metric as field in CheckResult

gvso:
The two new reports look like:

```yaml
rollout.cloud.run/lastHealthReport: |-
      status: healthy, but no enough time since last rollout
      metrics:
      - error-rate-percent: 0.00 (needs 1.00)
      - request-count: 5 (needs 0)
      lastUpdate: 2020-08-18T15:10:16-04:00
rollout.cloud.run/lastRollout: '2020-08-18T15:09:35-04:00'
```

```yaml
rollout.cloud.run/lastHealthReport: |-
      status: inconclusive
      metrics:
      - request-count: 0 (needs 100)
      lastUpdate: 2020-08-18T15:35:00-04:00
```

ahmetb:
🤔 it's sort of nice to omit other metrics in the status report when they're not relevant, but I suspect that might cause more confusion than it's helping. (e.g. people might think the metrics they configured aren't being taken into consideration by the program).

gvso:
I had a dilemma between your point and the user not knowing what `inconclusive` means when the other criteria seems fine. Maybe we should do `inconclusive, no enough requests`?

internal/health/health.go
```diff
@@ -82,16 +82,20 @@ func Diagnose(ctx context.Context, healthCriteria []config.HealthCriterion, actu
 		}
 
 		isMet := isCriteriaMet(criteria.Metric, criteria.Threshold, value)
+		result := CheckResult{Threshold: criteria.Threshold, ActualValue: value}
 
 		// For unmet request count, return inconclusive and empty results.
 		if !isMet && criteria.Metric == config.RequestCountMetricsCheck {
 			logger.Debug("unmet criterion")
```
ahmetb, line 89: duplicate log line

internal/health/health.go
```diff
@@ -82,16 +82,20 @@ func Diagnose(ctx context.Context, healthCriteria []config.HealthCriterion, actu
...
 			diagnosis = Inconclusive
-			results = nil
+			// TODO: This will be an issue if the request count is not the first
+			// criterion in the array since the caller can assume the order
+			// of the check results is the same as the order of the health
+			// criteria.
+			results = []CheckResult{result}
 			break
```
ahmetb, line 96: return all CheckResults here.
if there's unhealthy despite low req count, that's still unhealthy (+add a test case for this)

internal/rollout/rollout_test.go
```diff
@@ -293,7 +305,15 @@ func TestUpdateService(t *testing.T) {
 			lastReady: "test-002",
 			healthCriteria: []config.HealthCriterion{
 				{Metric: config.RequestCountMetricsCheck, Threshold: 1500},
-				{Metric: config.ErrorRateMetricsCheck, Threshold: 0.95},
+				{Metric: config.ErrorRateMetricsCheck, Threshold: 5},
+			},
+			outAnnotations: map[string]string{
+				rollout.StableRevisionAnnotation:    "test-001",
+				rollout.CandidateRevisionAnnotation: "test-002",
+				rollout.LastHealthReportAnnotation: "status: inconclusive\n" +
```
ahmetb, line 313: potentially it's gonna be a hassle to provide full values here.

internal/health/health.go
gvso: I gave `Unhealthy` precedence over `Inconclusive` and appended all results

internal/rollout/rollout_test.go
gvso: Yeah, let's work on this in another PR

PR #105, opened 2020-08-20
config: Support for Cloud Run on Anthos targets

gvso:
This adds flags to target services deployed to Cloud Run on Anthos. It adds fields to the Target struct and different constructors for managed and gke platforms


cmd/operator/main.go
```diff
@@ -91,8 +101,12 @@ func init() {
 	flag.BoolVar(&flCLI, "cli", false, "run as CLI application to manage rollout in intervals")
 	flag.DurationVar(&flCLILoopInterval, "cli-run-interval", 60*time.Second, "the time between each rollout process (in seconds)")
 	flag.StringVar(&flHTTPAddr, "http-addr", defaultAddr, "address where to listen to http requests (e.g. :8080)")
+	flag.StringVar(&flPlatform, "platform", config.PlatformManaged, "target platform to query for Knative services")
```
ahmetb: list valid values

internal/config/config.go
```diff
@@ -27,8 +33,18 @@ const (
 // }
 type Target struct {
 	Project       string
-	Regions       []string
 	LabelSelector string
+
+	// Either "managed" or "gke".
```
ahmetb, line 38: IMO in the long run we need to abstract this more and do something like:

```
type KServiceProvider interface {
	Services(ls labelSelector) ([]runapi.Service,error)
}
```

and we'd have GKEProvider, CloudRunManagedProvider.

Then they'd have their own built-in filters (e.g. cluster name, or regions) given at the time of initialization.

Otherwise any platform we add will overcrowd this Target type.

gvso: I have abstracted most things into a `knative` package, but the abstraction I currently have happens after the target had been set. For any other Knative provider, we might just need an endpoint to the cluster. Thought I'd would think more about how to abstract the target too.

cmd/operator/main.go
gvso: Done!

internal/config/config.go
gvso: As we discussed, we also need to consider authentication and other things, so this will be more complex. Let's work on this in another PR though

gvso:
@ahmetb I created a new branch, so we can push smaller changes without breaking the main one (there are changes needed in different components).

cmd/operator/main.go
```diff
@@ -91,8 +101,12 @@ func init() {
 	flag.BoolVar(&flCLI, "cli", false, "run as CLI application to manage rollout in intervals")
 	flag.DurationVar(&flCLILoopInterval, "cli-run-interval", 60*time.Second, "the time between each rollout process (in seconds)")
 	flag.StringVar(&flHTTPAddr, "http-addr", defaultAddr, "address where to listen to http requests (e.g. :8080)")
+	flag.StringVar(&flPlatform, "platform", config.PlatformManaged, "target platform to query for Knative services (managed or gke)")
 	flag.StringVar(&flProject, "project", "", "project in which the service is deployed")
 	flag.StringVar(&flLabelSelector, "label", "rollout-strategy=gradual", "filter services based on a label (e.g. team=backend)")
+	flag.StringVar(&flGKEClusterLocation, "cluster-location", "", "zone in which the cluster is located")
+	flag.StringVar(&flGKECluster, "cluster", "", "ID of the cluster")
```
ahmetb: might be prefix these descriptions with `(-platform=gke only)`

cmd/operator/main.go
```diff
@@ -91,8 +101,12 @@ func init() {
...
+	flag.StringVar(&flGKENamespace, "namespace", "default", "Kubernetes namespace where to look for Knative services")
```
ahmetb: if you have time, it would be good to implement "all namespaces" and not have this flag at all.

gvso: I'll try this out. The way I structured #107 should make it feasible to make this happen

PR #107, opened 2020-08-20
knative: Add knative package

gvso:
This adds a knative package and moves the run package inside it. The package declares an interface representing a Knative Provider, which the cloudrun package implements.

PR #108, opened 2020-08-24
Anthos stackdriver

gvso:
This adds a method to change the stackdriver query's fields to the ones needed by Cloud Run for Anthos. It uses an interface `platform` that returns the metric type for request latencies and request count needed for each case. For instance, CRFM uses `run.googleapis.com/request_latencies` while CRfA uses `knative.dev/serving/revision/request_latencies`


gvso:
This depends on #107 and #105

cmd/operator/main.go
```diff
@@ -59,10 +59,20 @@ var (
 	flCLI             bool
```
gvso, line 59: Ignore, same as #105

cmd/operator/rollout.go
```diff
@@ -10,14 +10,15 @@ import (
 	"github.com/GoogleCloudPlatform/cloud-run-release-manager/internal/metrics/sheets"
```
gvso, line 10: partly the same to #107

cmd/operator/service.go
```diff
@@ -5,8 +5,9 @@ import (
 	"sync"
```
gvso, line 5: Ignore, same as #107

internal/config/config.go
```diff
@@ -16,6 +16,12 @@ const (
 	ErrorRateMetricsCheck    MetricsCheck = "error-rate-percent"
```
gvso, line 16: Ignore, same as #105

internal/config/config_test.go
```diff
@@ -20,7 +20,7 @@ func TestStrategy_Validate(t *testing.T) {
 	}{
```
gvso, line 20: Ignore, same as #105

PR #109, opened 2020-08-26
rollout: Publish rollout events to PubSub

### LINKED ISSUE #75: Publish to Pub/Sub after a rollout/rollback have occurred

gvso:
This integrates the rollout package with the pubsub package to publish events after rollouts/rollbacks occur. It also adds a new flag `--notify-pubsub` to specify the PubSub topic.

### DIFF
--- a/internal/rollout/rollout.go
+++ b/internal/rollout/rollout.go
@@ -187,7 +187,7 @@ func (r *Rollout) replaceServiceAndPublish(svc *run.Service, trafficChanged bool
 	svc, err := r.runClient.ReplaceService(r.project, r.serviceName, svc)
 
 	if trafficChanged {
-		pubErr := r.publish(svc, diagnosis)
+		pubErr := r.publishEvent(svc, diagnosis)
 		if pubErr != nil {
 			r.log.Warnf("failed to publish rollout/rollback message: %v", pubErr)
 		}
@@ -249,7 +249,7 @@ func (r *Rollout) diagnoseCandidate(candidate string, healthCriteria []config.He
 	return d, errors.Wrap(err, "failed to diagnose candidate's health")
 }
 
-func (r *Rollout) publish(svc *run.Service, diagnosis health.DiagnosisResult) error {
+func (r *Rollout) publishEvent(svc *run.Service, diagnosis health.DiagnosisResult) error {
 	if r.pubsubClient == nil {
 		return nil
 	}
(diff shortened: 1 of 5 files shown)


internal/rollout/rollout.go
```diff
@@ -232,3 +248,23 @@ func (r *Rollout) diagnoseCandidate(candidate string, healthCriteria []config.He
...
+		return errors.Wrap(err, "failed to create message")
+	}
+	ctx := util.ContextWithLogger(r.ctx, r.log)
+	err = r.pubsubClient.Publish(ctx, event)
+	if err != nil {
+		return errors.Wrap(err, "error when publishing message")
+	}
+
+	// Wait for all messages to be sent (or to fail).
+	r.pubsubClient.Stop()
```
gvso: Since we're publishing **one** event per go routine (in each rollout process), we might just wait for the event to be published using [`PublishResult.Get`](https://pkg.go.dev/cloud.google.com/go/pubsub?tab=doc#PublishResult.Get). In this way, we can abstract the synchronous publishing from the user of the pubsub package

ahmetb: yeah that's a decent idea. I think that could be done with


    waitFn, err := PublishMessage()

internal/rollout/rollout.go
```diff
@@ -232,3 +248,23 @@ func (r *Rollout) diagnoseCandidate(candidate string, healthCriteria []config.He
 	d, err = health.Diagnose(ctx, healthCriteria, metricsValues)
 	return d, errors.Wrap(err, "failed to diagnose candidate's health")
 }
+
+func (r *Rollout) publish(svc *run.Service, diagnosis health.DiagnosisResult) error {
```
ahmetb: better name like publishEvent

internal/rollout/rollout.go
ahmetb: lets move Get into Publish

